PR #354, opened 2026-07-17
feat(app): experimental multi-window support ("New window")

ben-milanko:
## Summary

Adds a **"New window"** action to the DartPDF app that opens a second top-level OS window, built on Flutter's *experimental* desktop windowing API (`RegularWindowController` / `WindowManager` / `runWidget`, Flutter 3.44.x).

This is an **opt-in experiment gated behind the `windowing` feature flag** — released builds are untouched and stay single-window. When the flag is off (the default, and the test harness), everything degrades to today's `runApp` single-window behaviour, and the "New window" affordance is hidden.

> The windowing API is `@internal` and its own header warns *"Do not import this file in production applications … Flutter will make breaking changes, even in patch versions."* That's why it's flag-gated and quarantined to one file. Requested explicitly for experimentation on this branch.

### Turning it on

```
fvm flutter config --enable-windowing        # once; persists in config
fvm flutter run -d macos                       # or windows / linux
```

or inline: `--dart-define=FLUTTER_ENABLED_FEATURE_FLAGS=windowing`.

### What changed

- **`app/lib/window_support.dart`** (new) — the *only* file that touches the internal windowing API (file-level ignores for `invalid_use_of_internal_member` / `implementation_imports`). Small safe surface so nothing else imports `src/`: `multiWindowSupported`, `runDartPdfApp(root)` bootstrap (flag on → `runWidget(WindowManager(child: RegularWindow(...)))`, else `runApp`), and `openRegularWindow(...)` which registers a new window on the shared `WindowRegistry`.
- **`app/lib/app.dart`** — `DartPdfEditorApp` owns one shared `PdfEditingPreferences` and renders a reusable `_DartPdfWindow` (themed `MaterialApp` + `EditorScreen`). "New window" reuses the same prefs, so theme/tool styles stay in lock-step across windows. Only wired when `multiWindowSupported`.
- **`app/lib/main.dart`** — bootstrap routed through `runDartPdfApp`.
- **`app/lib/editor_screen.dart`** — optional `onNewWindow(BuildContext)` (null hides the affordance) and `persistSession` (default true). Adds app-menu item `menu-new-window` with a **⇧⌘/Ctrl+Shift+N** accelerator (plain ⌘N was already "New document").

### Session ownership

Each `EditorScreen` owns a `SessionStore`; letting every window restore + persist would double-open tabs and race the shared prefs key. Only the primary window keeps `persistSession: true` — secondary windows pass `false`, so they start empty and never write the store.

## Affected packages

- [x] Example app *(the `app/` DartPDF application)*
- [x] Documentation / CI / repository metadata only *(dev-log note)*

## Change type

- [x] Feature

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze` (root `app` — no issues)
- [x] Example app: `cd app && fvm flutter test` — all 176 tests pass (5 new)

The windowing flag is off in the test harness (no real OS windows there), so `app/test/multi_window_test.dart` pins the app-side wiring via the injectable seams: the `menu-new-window` item appears iff `onNewWindow` is provided and routes to it with a live context; ⇧⌘N and Ctrl+Shift+N both fire it; and a `persistSession: false` screen leaves a seeded stored session untouched after opening a document.

## Compatibility checklist

- [x] No `dart:ui` or Flutter imports outside `packages/dart_pdf_editor` — change is confined to the `app/` application, which is already a Flutter target.
- [x] No `dart:io` imports added in any `lib/` (the windowing bootstrap stays `dart:io`-free / web-safe).
- [x] Layering preserved — engine packages untouched.

## Notes for reviewers

Experimental / flag-gated; not intended for release builds. Known rough edges left for follow-up (called out in `doc/dev-log/2026-07-17-multi-window-experimental.md`):

- Every window's `EditorScreen` starts its own `IncomingFileService`, so an OS file-open/share can be delivered to *all* open windows.
- Each window registers a `WidgetsBindingObserver`, so the unsaved-changes exit guard runs per window.
- Closing the primary window doesn't force-quit while secondary windows remain (no `exit(0)` override — kept out to stay `dart:io`-free).

https://claude.ai/code/session_016AyAauDto2w739uQWzDxnn

---
_Generated by [Claude Code](https://claude.ai/code/session_016AyAauDto2w739uQWzDxnn)_

---

## Status (investigated 2026-07-25): merged up to main, but **cannot run yet** — two blockers

Merged `main` into the branch (97 commits behind); conflicts resolved, CI green,
330/330 app tests pass. Then tried to actually exercise it. It can't be, yet —
and the reasons are worth recording before anyone else spends the time.

### 1. Stable channel can never show the affordances

`onNewWindow` is null unless `multiWindowSupported`, which is Flutter's
`isWindowingEnabled` = `const String.fromEnvironment('FLUTTER_ENABLED_FEATURE_FLAGS')`
— a **compile-time** const. The tool only injects `windowing` when the feature is
*available on the channel*, and `windowingFeature` (flutter_tools
`features.dart`) declares `master: FeatureChannelSetting(available: true)` with
nothing for beta/stable; unspecified channels default to `available: false`.
`isEnabled` returns early on availability **before** reading any setting, so on
the stable SDK this repo pins (`.fvmrc`), all three routes are dead ends:

| Route | Result |
|---|---|
| `flutter config --enable-windowing` | value is written, never takes effect |
| `--dart-define=FLUTTER_ENABLED_FEATURE_FLAGS=…` | rejected outright by the tool |
| `FLUTTER_WINDOWING=…` env override | same availability gate |

So on stable the "New window" item is hidden **by design, not broken**. (For the
record: it lives in the **App menu**, 2nd item under "New document", ⇧⌘N — not
the native macOS menu bar — plus right-click a tab → "Move to new window".)

### 2. On master it doesn't compile, and once ported it crashes at launch

Tested against master **3.47.0-1.0.pre-221**:

**(a) Compile.** `window_support.dart` fails with 8 errors — the `@internal` API
was renamed and reshaped after this branch was written (**within 8 days**):

| | stable 3.44.8 | master 3.47.0 |
|---|---|---|
| controller | `RegularWindowController(preferredSize:)` | `WindowController(size:)` |
| widget | `RegularWindow` | `Window` |
| delegate | `RegularWindowControllerDelegate` | `WindowControllerDelegate` |
| manager | `WindowManager({child})` | `WindowManager({initialWindows})` |

The last is structural: master's `WindowManager` renders each registered entry
itself, so the primary window becomes the first `WindowEntry`. The two shapes are
**mutually exclusive** — porting would break the build on stable and turn this
PR's CI red, which is why no port is included here.

**(b) Launch.** Ported (4 edits, analyzes clean on master), it builds and then
aborts immediately:

```
NSInternalInconsistencyException: 'Multiview can only be enabled before adding
any view controllers.'
  -[FlutterEngine enableMultiView]
  InternalFlutter_WindowController_CreateRegularWindow
```

`WindowController(...)` asks the engine to enable multiview, but
`macos/Runner/MainFlutterWindow.swift`'s `awakeFromNib()` has already created a
`FlutterViewController` and set it as `contentViewController`. This PR contains
**no native changes** (its commits touch 5 Dart files, `codecov.yml` and a
dev-log), so **multi-window has never actually run** — the 9 tests here cover the
Dart-side wiring (`onNewWindow`, handoff, `persistSession`) and nothing more.

Making it run means reworking the macOS embedding: no NIB-created view
controller, own a `FlutterEngine` directly, and rehost all five method channels
(`incoming`, `file_access`, `image_clipboard`, `native_print`) plus
`RegisterGeneratedPlugins` on that engine. Those channels are load-bearing
(security-scoped bookmarks, ranged reads, printing, clipboard), so it's a
redesign rather than a tweak — and it could not merge while `.fvmrc` pins stable.

### Recommendation

**Park it.** The API broke in 8 days, so anything built against it now will
likely break again before it can ship. Keep this as reviewed Dart-side wiring
with green CI, and revisit when Flutter promotes windowing to beta/stable — at
which point both the API port and the native rework should be done together.

The misleading setup instructions in `window_support.dart` (which recommended
the rejected `--dart-define` recipe and never mentioned the channel requirement)
are corrected in this branch.


ben-milanko:
Superseded by #680, which rebuilds the feature against Flutter 3.47's current windowing API and native runner lifecycle.

Issue #314, opened 2026-07-16
perf: GPU-resident PdfTileStore — composite budgeted tiles instead of re-rasterizing whole pages


ben-milanko:
From the 2026-07-16 architecture review (perf focus), added as candidate 9 on request. This is the structural gap against native viewers (Bluebeam-class): they composite GPU-resident tiles and never re-rasterize on pan; we re-read the whole page from the GPU on every settle.

## Current shape (verified against source)
Three coexisting raster mechanisms per `_PdfPageViewState`, each with its own staleness rule:
- **Full-page `_image`** — one `picture.toImage` per 200 ms settle (`pdf_page_view.dart:1346–1372`), even when 10% of the page is visible; capped by `_maxPixels = 1<<24` / `_maxDimension = 8192` (`:380–388`), so page size bounds fidelity.
- **Exactly one detail patch** — `_detailImage`/`_detailFraction` (`:369–370`), computed by `_detailGeometryAt` (`:1695`), **no budget, no cache**: it dies on every settle, scene change, and edit, so panning past it means blur until a fresh region `toImage` lands.
- **`_preview` upscale** as the third fallback layer.

Per-page retention (`_scene`, `_image`, `_slugPicture`, `_detailImage`) is bounded only by the lazy-list build window — part of #283's growth. During gestures nothing re-rasters (good — stale raster scales under the InteractiveViewer transform), but the settle pays a whole-page GPU readback, which on web is single-threaded and uncancellable (`pdf_page_view.dart:363–366`).

## Proposed module
`PdfTileStore` — a quantized zoom-bucket pyramid of fixed-size GPU-resident tiles (`ui.Image`, 512² physical px ≈ 1 MB RGBA).

**Interface (small):**
- `TileView viewFor(PdfPageRenderIntent intent, PdfPageRenderPlan plan, double zoomBucket, Rect visiblePageRect)` — sync, returns best-available tiles now (falling back per-tile to the nearest coarser bucket, upscaled)
- `invalidate({Set<int>? pages})`
- a `Listenable` that ticks as sharper tiles land; the page view repaints tiles via `canvas.drawImageRect`

[...]

PR #365, opened 2026-07-18
fix(#360): devtools tile counters read the effective store, not just the override

ben-milanko:
## Problem

The F12 devtools snapshot (both the JSON export and the live "Deep-zoom detail (#314)" panel) read the tile store from `PdfPageView.debugTileStoreOverride`, which is **only non-null when someone manually flips the deep-zoom mode in devtools** (that path spawns a fresh override `PdfTileStore`).

On the **default desktop path**, tiles are served from `PdfTileStore.instance` and the override is `null` — so the entire `tileStore` counter block (`scheduled` / `landed` / `discarded` / `batches`) was **omitted from the snapshot on the exact platform where the #314 tile pyramid ships on-by-default.**

This is a diagnostics blind spot: while investigating tile eviction-thrash / green-grid flicker on macOS release (198-page CAD book, tiles cache pegged at its 100 MB budget, every frame janking), the on-device traces couldn't show the real tile-store churn counters, because the panel was reading the wrong (null) store. The tiles *cache* row survived only because it comes from a separate cache registry.

## Fix

Read the *effective* store — mirroring the render path (`debugTileStoreOverride ?? PdfTileStore.instance`) — but only when the tile path is actually enabled:

```dart
final store = PdfPageView.tileStoreDetail
    ? (PdfPageView.debugTileStoreOverride ?? PdfTileStore.instance)
    : PdfPageView.debugTileStoreOverride;
```

In `patch` mode (`tileStoreDetail == false`) the ternary yields `null`, so the existing `if (store != null)` guard still omits the block — no misleading empty block, and no needless lazy instantiation of the singleton. Applied in both `_exportSnapshot` and `_deepZoomSection`.

Diagnostics-only change; no behavior change to the tile pyramid itself.

## Verification

- `fvm dart analyze` (root): **No issues found**
- `app/test/devtools_panel_test.dart` + `devtools_options_test.dart`: **7/7 pass**

## Context

This unblocks confirming/quantifying the tile eviction-thrash on the default desktop path (issues #314 / #360). Root-cause fix for the thrash itself (a `viewFor` budget-vs-demand guard + intra-frame eviction pin) is a separate follow-up.

### DIFF
--- a/app/lib/devtools_panel.dart
+++ b/app/lib/devtools_panel.dart
@@ -137,7 +137,9 @@ class _DevToolsPanelState extends State<DevToolsPanel> {
   Future<void> _exportSnapshot() async {
     final frames = _tools.frameStats();
     final session = widget.session;
-    final store = PdfPageView.debugTileStoreOverride;
+    final store = PdfPageView.tileStoreDetail
+        ? (PdfPageView.debugTileStoreOverride ?? PdfTileStore.instance)
+        : PdfPageView.debugTileStoreOverride;
     final snapshot = <String, Object?>{
       'tool': 'dartpdf-devtools',
       'exportedAt': DateTime.now().toIso8601String(),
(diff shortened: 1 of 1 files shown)


ben-milanko:
Superseded by #371, which folds this exact diagnostic commit (devtools tile counters read the effective store) in alongside the tile eviction-thrash fix it was written to help confirm. Closing to consolidate into one PR — no changes lost.

Issue #364, opened 2026-07-18
perf: progressive open for mobile cloud picks (OneDrive/iCloud) — bypass the picker's whole-file copy


ben-milanko:
Follow-up to #359 / PR #363. #359 made the desktop open path progressive (first paint from ranged reads, full bytes streamed in behind it). **Mobile is not covered**, and it's a real gap — not "already in memory".

## Why mobile isn't covered by #363

You can pick a OneDrive/iCloud/Drive file on a phone, and it *does* pay the full cloud transport — but that cost is paid **inside the OS document picker's copy**, before the app sees a byte or a path, so app-level ranged open (the #359 mechanism) can't intercept it:

- **Android** — `file_selector_android`'s `FileUtils.getPathFromCopyOfFileFromUri` does `contentResolver.openInputStream(uri)` → `copy(...)` streaming the **entire** content URI into `{cacheDir}/{uuid}/{fileName}`, then returns that local path. `readAsBytes` afterwards just reads the local copy.
- **iOS** — `file_selector_ios` uses `UIDocumentPickerViewController(documentTypes:in: .import)`, which **copies** the file into a temp location.

By the time the app could open anything, the whole file is already local. So on `progressiveOpenSupported` we gate mobile off (it only returns true for a desktop path).

Mitigating detail: mobile **reopens** (recents / session restore) already read the local `cacheOpenedPdf` snapshot, so they're fast. Only the **first pick** of a cloud file pays transport.

## What a mobile solution needs

1. **A custom picker that keeps a reference to the original, not a copy:**
   - Android: `ACTION_OPEN_DOCUMENT` (via a platform channel), hold the returned `content://` `Uri` (and persist read permission with `takePersistableUriPermission` for reopens).
   - iOS: `UIDocumentPickerViewController(forOpeningContentTypes:)` in `.open` mode (not `.import`), keep the security-scoped URL.
2. **A mobile `PdfByteSource` doing native ranged reads over that reference:**
   - Android: `ContentResolver.openFileDescriptor(uri, "r")` → seek + read a range. **Risk/unknown:** many cloud providers (incl. OneDrive) return a **non-seekable** `ParcelFileDescriptor` (a pipe) from SAF, so random access may be impossible for exactly the providers we care about — in which case ranged reads fall back to a full stream and there's no win. This must be probed per provider before committing.
[...]

Issue #359, opened 2026-07-18
perf: open local files through the ranged byte source — first paint without reading the whole file


ben-milanko:
#325 gave us the asynchronous random-access byte-source API (PR #328: `PdfByteSource`, `CosDocument.openSource`, `PdfDocument.openSource`, `PdfHttpByteSource` with Range/fallback). This issue is the missing half: **the app's own open path never uses it**. Every open — File menu, drag-drop, recents, session restore — slurps the complete file into memory (`readAsBytes`) before `PdfDocument.open` sees a byte.

## Why it matters (measured 2026-07-18)

A 198-page scanned test book (179 MB) living in OneDrive Files-On-Demand:

| step | time |
|---|---|
| raw read of the file through the macOS file provider (`cat`, 0% CPU) | **~36 s, every open** (not just first hydration; ~5 MB/s) |
| full parse of the same bytes from local disk | 1.5 s |
| per-page render, headless debug | ~1 s |

So the user experience is 30–40 s of "Opening…" that is 100% byte transport. Cloud-synced folders (OneDrive/Dropbox/iCloud) are the *normal* home of this app's target documents — big scanned test books and CAD sheets — so this is the common case, not an edge case.

## Proposal

A `PdfFileByteSource` (random-access reads over a `RandomAccessFile`; trivially satisfies the `PdfByteSource` contract) plus wiring the app's open path through the progressive loader:

1. **First paint from ranges**: the progressive loader already fetches the header probe, xref chain, and live-object ranges into a sparse buffer — on a file source that is a few MB of reads, so the viewer opens in ~1–2 s even on a slow provider.
2. **Full buffer in the background**: the editing session (incremental save appends to one grow-only buffer), signature validation over full byte ranges, and the render workers all want complete bytes today — so kick off the sequential read behind the first paint and swap/complete when it lands (the loader's sequential-download fallback is exactly this shape). Edit affordances that need the full buffer (Save, Digitally sign) stay disabled or queue until then.
3. **Progress**: `PdfHttpByteSource` already models progress + cancellation; the file source should feed the same reporting so the "Opening…" placeholder can show real progress for the background full read.

Constraints/notes:
- Worker isolates each get a copy of the document bytes; on a 179 MB doc the spawn copies are ~500 MB across the pool — worth revisiting spawn-time transfer (TransferableTypedData or the shared-source model) in the same effort, since progressive open makes big-doc sessions more common.
[...]

PR #372, opened 2026-07-18
perf(#364): progressive open for mobile cloud picks — reference picker + native ranged reads

ben-milanko:
## Summary

Follow-up to #359 / #363. #363 made the **desktop** open path progressive (first paint from ranged reads, full bytes streamed in behind it) but gated **mobile** off: a phone pick of a OneDrive/iCloud/Drive file pays its whole cloud transport *inside the OS picker's copy*, before the app sees a byte or a path, so app-level ranged open couldn't intercept it.

This adds a **reference picker** + a **mobile ranged `PdfByteSource`**, wired into the same `_openProgressive` swap machinery #363 built. **Desktop is untouched.**

- **`PdfMobileByteSource`** (`app/lib/pdf_mobile_source.dart`) — reads ranges over an opaque native token through the `dev.milanko.dartpdf/mobile_file` method channel; the mobile sibling of `PdfBookmarkFileByteSource`. `file_io` gains `MobilePickedPdf`, `supportsMobileProgressiveOpen`, `pickPdfMobileReferences`, `pdfByteSourceForMobileToken`.
- **Wiring** (`editor_screen.dart`) — `_openProgressive` now takes `path` **or** `token`; `_progressiveSource` / `_readOriginFully` pick the origin; `DocumentTab` carries `originToken`. `_pickAndOpenMobile` opens a single **seekable** pick progressively and streams a **non-seekable / batch** pick whole (one read of the original, no OS copy — never worse than the old path). The complete bytes are snapshotted on the full-read swap so Recent/session restore reopen from the local copy. Falls back to the copy-based picker on a runner without the channel.
- **Android** (`MainActivity.kt`) — `ACTION_OPEN_DOCUMENT` keeping the `content://` Uri (persisted read permission) + `openFileDescriptor` / `Os.lseek` ranged reads. A `statSize` + `lseek` **seekability probe** detects the non-seekable SAF pipe many cloud providers return and streams whole instead.
- **iOS** (`SceneDelegate.swift`) — `UIDocumentPickerViewController` in `.open` mode + a security-scoped bookmark token + `NSFileCoordinator` coordinated `FileHandle` ranged reads.

Fail-safe by construction: a missing channel, cancelled pick, non-seekable provider, or any first-paint failure falls back to a whole read, so nothing a plain mobile open handled can regress — the only variable is whether the first-paint speedup materializes per provider.

## Affected packages

- [x] Example app

## Change type

- [x] Feature
- [x] Performance change

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze` (app — clean)
- [x] `cd packages/dart_pdf_editor && fvm flutter test` (full `app/` suite green, 233 tests; new `pdf_mobile_source_test.dart` + `mobile_progressive_open_test.dart`)

**The native handlers cannot be exercised from CI or the desktop harness.** The Dart contract is covered by mocked-channel unit tests and an end-to-end pick → first-paint → full-read-swap widget test on an Android platform override. The Android/iOS ranged-read code itself needs **on-device verification against real cloud providers** — this is #364's stated acceptance criteria and is the pending work:

1. **Android spike:** confirm `openFileDescriptor` yields a **seekable** FD for a real OneDrive/Drive file on a device. If cloud providers return pipes, `probeSeekable` returns false and those picks stream whole (correct, no progressive win). If no mainstream provider is seekable, Android progressive isn't viable via SAF and should be scoped to iOS only — the plumbing already degrades to the stream path automatically.
2. **iOS:** verify a coordinated ranged read paints a large iCloud/OneDrive PDF before the whole file downloads.

## Compatibility checklist

- [x] No `dart:ui` or Flutter imports outside `packages/dart_pdf_editor`. (New Dart lives in the example `app/`, which already depends on Flutter.)
- [x] No `dart:io` imports in any `lib/` directory. (`PdfMobileByteSource` uses only the method channel.)
- [x] Layering is preserved.

## Notes for reviewers

- Design + the on-device verification gap are written up in `doc/dev-log/2026-07-18-mobile-progressive-open.md`.
- The whole feature degrades safely to the prior behavior on any failure path, so it can land ahead of the device spike; the spike decides whether Android keeps the progressive path or is scoped to iOS only.

---
_Generated by [Claude Code](https://claude.ai/code/session_018WRoDHgri4qhYcU9tbFvEr)_

### DIFF
--- a/app/android/app/src/main/kotlin/dev/milanko/dart_pdf_editor_app/MainActivity.kt
+++ b/app/android/app/src/main/kotlin/dev/milanko/dart_pdf_editor_app/MainActivity.kt
@@ -1,5 +1,6 @@
 package dev.milanko.dart_pdf_editor_app
 
+import android.app.Activity
 import android.content.ClipData
 import android.content.ClipboardManager
 import android.content.Context
@@ -14,11 +15,17 @@ import android.print.PrintDocumentAdapter
 import android.print.PrintDocumentInfo
 import android.print.PrintManager
 import android.provider.OpenableColumns
+import android.system.ErrnoException
+import android.system.Os
+import android.system.OsConstants
 import io.flutter.embedding.android.FlutterActivity
 import io.flutter.embedding.engine.FlutterEngine
+import io.flutter.plugin.common.MethodCall
 import io.flutter.plugin.common.MethodChannel
 import java.io.File
+import java.io.FileInputStream
 import java.io.FileOutputStream
+import java.util.concurrent.Executors
 
 /// Forwards PDFs the OS opens in the app - a Files "open", a download tap, or a
 /// share - to the Dart `IncomingFileService` over a single method channel.
@@ -26,10 +33,20 @@ class MainActivity : FlutterActivity() {
     private val channelName = "dev.milanko.dartpdf/incoming"
     private val imageClipboardChannelName = "dev.milanko.dartpdf/image_clipboard"
     private val nativePrintChannelName = "dev.milanko.dartpdf/native_print"
+    private val mobileFileChannelName = "dev.milanko.dartpdf/mobile_file"
     private var channel: MethodChannel? = null
 
     /// The file the activity was launched with, drained by `getInitialFile`.
     private var pending: Map<String, Any>? = null
+
+    /// Serializes the (potentially large) ranged FD reads off the platform
+    /// thread so a multi-MB read can't ANR the UI.
+    private val fileIoExecutor = Executors.newSingleThreadExecutor()
+
+    /// The in-flight `pickDocuments` reply, held while the system document
+    /// picker is up (its result arrives asynchronously in `onActivityResult`).
+    private var pendingPick: MethodChannel.Result? = null
+    private val openDocumentRequestCode = 0x0D_0C
 
     override fun configureFlutterEngine(flutterEngine: FlutterEngine) {
         super.configureFlutterEngine(flutterEngine)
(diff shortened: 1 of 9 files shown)


ben-milanko:
## On-device confirmation — Android + OneDrive ✅

Validated against real OneDrive files on an Android release build (devtools traces, 2026-07-19). The Android acceptance criterion — *does `openFileDescriptor` yield a **seekable** FD for a mainstream cloud provider?* — is **met for OneDrive**.

### Evidence (not inference)

`rangeRequests` / `rangeBytesFetched` are emitted only by the ranged `PdfByteSource` (`pdf_cos/src/byte_source.dart`). The pre-#372 mobile path copies the file and opens a whole in-memory buffer → **no ranged source, so these are 0**. They are non-zero on `platform: android`, so the new `PdfMobileByteSource` is the code path running.

Which ranged path fired discriminates seekable-vs-pipe:
- **Non-seekable SAF pipe** → `_downloadFully` (sequential) → `rangeBytesFetched ≈ whole file`.
- **Seekable FD** → the sparse `_fetch` loader reads only the byte gaps it needs at arbitrary offsets.

Observed: `rangeBytesFetched: 444 KB` across three OneDrive files of **58 / 21 / 36 MB** (`rangeRequests: 122`). 444 KB ≈ ~1% of a single 36 MB file — this can only be the **sparse seekable path**; a pipe would have shown ~115 MB here.

Progressive swap also confirmed end-to-end, per document:
```
first paint — "Westona…" 107 pages; reading full file…
full read complete — 36,383,879 bytes        (~636 ms later)
```
First paint from ranged reads (~148 KB/doc), full bytes streamed behind. `objectRescueScanBuilt: 1` — one doc had a broken xref and recovery still worked over the ranged source.

### Still pending for full acceptance
- [ ] **iOS** — coordinated ranged read paints a large iCloud/OneDrive PDF before the whole file downloads (not exercised; these are Android-only traces).
- [ ] **Other providers** — only OneDrive confirmed. Google Drive / iCloud outstanding; a pipe-returning provider will degrade to whole-stream (correct, but no win).
- [ ] **Speedup magnitude** — logs prove progressive *works* and fetches ~1% for first paint, but not that first paint is *faster* than the old copy path (no A/B). `sourceFetch: 4.87 s (1 call)` is a slow-transport signal, but that's the full-file backfill stream, not the tiny ranged first-paint path.

**Bottom line:** the Android/SAF seekability gamble pays off on OneDrive, and the safe-degradation plumbing behaves as designed. Android progressive is viable (not iOS-only). Remaining work is the iOS capture + a Drive/iCloud provider check.

PR #388, opened 2026-07-19
perf(editor): X-strip transcript banding to bound retained memory on extreme pages (#385)

ben-milanko:
## Summary

Follow-up to #383 (grid-cull region replay). On a single flat-content-stream page the retained `PdfRenderCommand` transcript is an irreducible memory floor (~60–100 MB on the 8504×842 CAD sheet). #383's spatial cull skips *replaying* offscreen commands but still *retains* all of them. Because such pages are extreme-aspect and panning is horizontal, this partitions the page into N vertical strips along the pan (X) axis and retains only the strips overlapping the viewport, bounding retained memory to roughly `visible strips / total`.

This lands the tested **engine + primitives + diagnostics + memory-pressure hook**. Full viewer adoption (the page view dropping the flat transcript for extreme pages and driving eviction off the scroll position) is deliberately a follow-up — this is the mechanism it sits on.

- **`PdfBandedTranscript`** (`banded_transcript.dart`) — the retention/eviction engine on the same supported subset `PdfRegionReplayIndex` already guarantees (balanced save/restore, nestable clips, no transparency/soft-mask group split). A strip-spanning unit is retained by *every* strip it overlaps (shared command object, not a copy), so evicting the others never drops a paint a survivor still needs; `replayRegion` dedupes such a unit back to a single draw in painter order — **byte-identical** to whole-transcript region replay. `evict` / `evictOutside(keep)` free strips, `estimatedResidentBytes` (identity-deduped) tracks the distinct paints actually held, `missingBandsForRegion` detects a pan into a dropped strip, and `restore(freshCommands)` re-materializes from one re-record (the ~760 ms full-page record the decomposition trades against; `restore` only re-partitions it).
- **`PdfRegionReplayIndex.unitBandHistogram(bands)`** + `xExtent` — the reintroduced `debugUnitBandHistogram` diagnostic quantifying the uneven per-strip unit spread (the CAD probe's densest 3 of 10 strips carried ~54%).
- **`PdfRetainedScene`** — `bandTranscript` / `bands` / `reband`, `dropRegionIndex` (the reintroduced primitive — frees the index and any banded transcript, both rebuild identically, the authoritative `commands` stay intact), `debugUnitBandHistogram`, and a weak live-scene registry backing `PdfRetainedScene.handleMemoryPressure`, now driven by the viewer's `didHaveMemoryPressure`. Dropped indices rebuild identically and evicted strips re-materialize, so there is **no visual regression**.

## Affected packages

[...]

Issue #384, opened 2026-07-19
perf(editor): build the region-replay grid index on the render-worker isolate


ben-milanko:
Follow-up to #383 (grid-cull region replay).

## Context
With the grid index (#383), a dense CAD page's deep-zoom tiles are cheap. The remaining UI-isolate stall is the **grid index build itself** — walking the ~850k-command transcript to bin ~282k paint units. Measured at **~210 ms** on first deep-zoom, on the UI isolate.

That build is **pure data** (command-bounds arithmetic, no `dart:ui`), so it can move to the render-worker isolate, which already produces the command transcript via `worker.record(...)`. This removes the last first-zoom freeze on heavy pages.

## Work
- Build `PdfRegionReplayIndex` on the worker alongside `record`, or as a separate worker request keyed by page.
- Serialize it back across the isolate boundary. The grid arrays (`_cellStart`, `_cellUnits`, `_broad`, unit bounds/commandIndex/blendMode) are trivially sendable as typed arrays; the clip-state tree (`PdfRegionClipState`, holding `PdfClipPathCommand`/`PdfPath`) needs a serialization path — reuse or extend the existing `serializeCommands`/`deserializeCommands` machinery.
- Wire `_ensureRegionIndex()` to accept a worker-supplied index and fall back to the in-isolate build when the worker declines/isn't active.

## Prototype available
The session that produced #383 prototyped an in-isolate `PdfRetainedScene.warmRegionIndex()` (build ahead of the gesture) and a `dropRegionIndex()` memory-pressure primitive, both tested byte-identical. They were left out of #383 to keep it scoped; they're the natural landing spot for this work.

## Acceptance
- First deep-zoom on the CAD probe page shows no ~210 ms UI-isolate build in a trace.
- Region rasters remain byte-identical to the in-isolate build.

ben-milanko:
Landed in #389 (`3e694de5`). `PdfRenderWorker.buildRegionIndex` + `PdfRetainedScene.warmRegionIndex` are on main (5 references in `render_worker.dart`), with `region_replay_index.dart` made Flutter-free so both the native isolate and the `dart compile js` web worker can build the index off the UI thread.

PR #389, opened 2026-07-19
perf(editor): build the region-replay grid index on the render-worker isolate

ben-milanko:
## Summary

Follow-up to #383 (grid-cull region replay). **Stacked on #383** — set the base to `perf/cad-region-grid-and-hover` so the diff is only this change; needs #383 to land first (or retarget to `main` once it merges).

With the grid index (#383), a dense CAD page's deep-zoom tiles are cheap. The remaining first-deep-zoom UI-isolate stall was the **grid index build itself** — walking the ~850k-command transcript to bin ~282k paint units, ~210 ms on the UI isolate. That build is **pure data** (command-bounds arithmetic, no `dart:ui`), so it moves to the render-worker isolate, which already produces the transcript via `worker.record`.

What changed:

- **`region_replay_index.dart` is now Flutter-free.** `replay` routed save/restore through a `dart:ui` `Canvas`; it now routes them through `PdfDevice` (whose `CanvasPdfDevice` forwards to the canvas — byte-identical). Dropping the `package:flutter/painting.dart` import lets **both** worker backends (the native isolate and the `dart compile js` web worker) import the file and call `PdfRegionReplayIndex.build`. Verified the web worker still compiles to JS.
- **`serializeRegionReplayIndex` / `deserializeRegionReplayIndex`.** The grid arrays (`_cellStart`, `_cellUnits`, `_broad`) and per-unit bounds/`commandIndex`/`blendMode` are trivially sendable typed data; the clip-state tree (`PdfRegionClipState` → `PdfClipPathCommand`/`PdfPath`) is flattened parent-before-child to index references, and the clip **path** commands ride the existing `serializeCommands`/`deserializeCommands` codec — the same float32 truncation the scene transcript already carries, so the round trip is pixel-lossless.
- **`PdfRenderWorker.buildRegionIndex(page, {annotations, maxCommands, buildGrid, priority})`** — base declines (`null`); native isolate + web worker override; pool routes by stable page affinity like `binStrips`; caching wrapper is a pure passthrough. Both backends build from the **same round-tripped, compacted wire transcript** the strip bins replay, so the index's unit indices line up with a worker-recorded scene's `commands`.
- **`PdfRetainedScene`**: `warmRegionIndex(worker)` (off-isolate build with in-isolate fallback, idempotent), `dropRegionIndex()` (memory-pressure primitive), `regionIndexBuildIsHeavy` (offload only the grid-escalation case), and `_ensureRegionIndex` sharing `_regionIndexBuildParams` with the worker request.
- **`pdf_page_view.dart`**: warm the index on the worker ahead of the tile path; a dense page defers the tile path until the warmed index is resident rather than paying the ~210 ms build synchronously. **Pages below the grid ceiling are unchanged** (cheap synchronous linear build).

## Affected packages

- [x] `dart_pdf_editor`

## Change type

- [x] Performance change

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze` (clean, all packages)
- [x] `cd packages/dart_pdf_editor && fvm flutter test` (new `region_replay_index_worker_test`; `region_replay_grid_test`, `retained_scene_test`, `spatial_region_corpus_test`, `render_worker_test`, `render_worker_strips_test`, `tile_store_page_view_test`, `strip_zoom_router_test`, `strip_plan_device_test`, `render_worker_transcript_cache_test` all green)
- [x] Web worker `dart compile js` compiles (proves the Flutter-free constraint)

Uses the existing checked-in `test_corpora/pdfjs/` fixtures — no new PDFs.

## Compatibility checklist

- [x] No `dart:ui` or Flutter imports outside `packages/dart_pdf_editor` (this change actually *removes* the Flutter import from `region_replay_index.dart`).
- [x] Layering is preserved.

## Notes for reviewers

- **Byte-identical proof.** The worker re-records through `_BinCommandCache`/`PdfWorkerTranscriptCache`, producing `serialize(compactStateScopes:true, decodeImages:false)` → `deserialize`; a worker-recorded scene's `commands` are `serialize(compactStateScopes:true, decodeImages:true)` → `deserialize`. `compactStateScopes` removes save/restore purely on clip structure and `decodeImages` only changes image payloads, so the two transcripts are structurally identical and the index's `commandIndex` positions align. `region_replay_index_worker_test` proves it end to end through a **real background-isolate** `buildRegionIndex` plus a serialize/deserialize round trip across a diverse corpus.
- **Acceptance.** First deep-zoom on a dense page no longer pays the ~210 ms build on the UI isolate (it warms on the worker; the tile path defers until the index lands). Region rasters stay byte-identical to the in-isolate build.
- **Worker-queue occupancy.** The grid build is synchronous (can't yield mid-build), so it occupies the worker queue for its duration; acceptable because the visible page's record already completed before its index is warmed, so only background prefetch / a later settle's bin waits.
- **`dropRegionIndex()`** is provided and tested here as the memory-pressure primitive issue #384 names as a landing spot (the #383 prototype). The viewer-side wiring into `didHaveMemoryPressure` is **tracked by #388** (its weak live-scene registry + `handleMemoryPressure`), not a follow-up on this PR — so it isn't implemented twice. ⚠️ #388 also introduces `PdfRetainedScene.dropRegionIndex`, so whichever of the two lands second should reconcile the shared method + memory-pressure surface rather than re-adding it (this branch is stacked on #383, #388 targets `main`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01ESxQkp9qteJWsorfqwwYH3)_


ben-milanko:
## Cross-reference from the performance-roadmap audit (#408)

Two notes from auditing the roadmap tickets against this branch.

**1. The unwired-`dropRegionIndex` follow-up is resolved by #388, not still open.** The reviewer note here says `dropRegionIndex()` is provided and tested but not wired into `didHaveMemoryPressure` because *"that handler doesn't enumerate per-page scenes"*. #388 adds exactly that — "a weak live-scene registry backing `PdfRetainedScene.handleMemoryPressure`, now driven by the viewer's `didHaveMemoryPressure`" — so the enumeration lands there. Worth striking the follow-up from this PR's notes and pointing at #388 so it doesn't get implemented twice.

**2. Possible collision with #388.** Both PRs introduce `PdfRetainedScene.dropRegionIndex` — it's already on this branch (`retained_scene.dart:370`) and #388 describes it as "the reintroduced primitive". #388 is currently `CONFLICTING` against `main`. Whichever lands second will need the other's version reconciled rather than re-added; the two also both touch `PdfRetainedScene`'s memory-pressure surface. Flagging before either is rebased.

I've also noted on #405 (global live-raster budget) that #388's live-scene registry plus `estimatedResidentBytes` is the retained-scene half of that ticket, which narrows #405 to registering full-page rasters and detail patches with the existing `PdfCacheRegistry`.

ben-milanko:
Thanks — updated the reviewer notes accordingly.

1. **Wiring follow-up struck.** The `dropRegionIndex()` note now points at #388 as the owner of the `didHaveMemoryPressure` wiring (live-scene registry + `handleMemoryPressure`) instead of describing it as an open follow-up here, so it won't get implemented twice.

2. **Collision.** Leaving `dropRegionIndex` in this PR since #384 explicitly names it as a landing spot and the byte-identical drop→rewarm test exercises it — but I've flagged in the notes that #388 also introduces it, so whoever lands second reconciles the shared method/memory-pressure surface rather than re-adding. This branch is stacked on #383 (`perf/cad-region-grid-and-hover`) and #388 targets `main`, so the reconciliation will happen at whichever rebase lands last; happy to be the one to absorb #388's version if it lands first.

---

Issue #310, opened 2026-07-16
pdf_graphics: extract a PdfColorSpace module (colorspace→sRGB derived 4 times, already diverged)


ben-milanko:
From the 2026-07-16 architecture review.

## Files
- `pdf_graphics/lib/src/interpreter.dart:99–322` — private `_PdfColorSpaceResolver`/`_PdfResolvedColorSpace` (ICC + calibrated + tint)
- `pdf_graphics/lib/src/image_pixels.dart:1529` — `_AlternateColorSpace`, `_isDirectDeviceColorSpace`, `pdfImageDecodeRanges`
- `pdf_graphics/lib/src/shading.dart:255–320` — `_componentCount`/`_colorConverterFor` (tint + calibrated, **no ICC**)
- `pdf_graphics/lib/src/mesh.dart`, `color.dart` (`colorFromComponents`)

## Problem
The knowledge "given a COS colorspace object, how many components does it have and how do N component values become sRGB" (DeviceGray/RGB/CMYK, CalGray/CalRGB/Lab, ICCBased/N, Indexed, Separation/DeviceN tint transforms) is re-derived independently in four modules, each handling a slightly different subset. The copies have already diverged: ICC is supported in the interpreter path and absent from shading.

## Solution
One `PdfColorSpace.parse(cos, csObject)` returning `int channels` + `PdfColor toSrgb(List<double>)` (plus a sample-decode helper for images). `IccProfile` and `PdfCalibratedColorSpace` become internal strategies; `colorFromComponents` the device-family fallback. Interpreter, shading, mesh, and image_pixels all consume the one interface.

## Wins
- Deletion test passes hard: delete it and the same Separation/DeviceN/ICC/Lab parsing reappears across four callers
- The ICC gap in shading closes for free
- First direct `color_space_test.dart` becomes possible — today conversion is only reachable through whole-page `interpreter_test`/`shading_test`
- Pulls the untested cluster (`function.dart` 528 lines, `mesh.dart` 376, `calibrated_color.dart` 325 — no dedicated tests) up to a testable interface

Strength: Strong.

PR #391, opened 2026-07-19
chore(pdfjs): capture the devicen.pdf render drift in the gallery

ben-milanko:
## Summary

⚠️ **This records a regression rather than refreshing a stale artifact.** Please read before merging.

`test_corpora/pdfjs/_renders/devicen.pdf.p0` went from a **0.000%** baseline diff to **3.785%**. This PR regenerates the gallery PNGs so they match what the renderer currently produces — it is *not* an endorsement of the new pixels.

The distinction matters: the page previously matched the PDF.js baseline **exactly**. The diff is two large solid triangles, which is a real rendering change, not antialiasing noise. Merging this makes the gallery self-consistent but also makes the regression invisible, so it may be preferable to hold this until the cause is fixed.

## Why it is not attributable to any open branch

- Reproduces identically (3.785%, same diff image) on `origin/main` with nothing applied.
- Reproduces on #386 and #387, neither of which touches DeviceN or colour handling.
- `devicen.pdf.p0.dart.png` has not been modified since `0b0b62c0` ("chore: check in pdfjs visual render artifacts").

So something merged after `0b0b62c0` changed DeviceN rendering without the gallery being refreshed.

## Environment control

This was regenerated on macOS with system fonts present. In the same full-suite run, **every other PDF in the corpus reproduced byte-identically** to the checked-in PNGs, which rules out a font/platform difference on my machine and isolates the change to devicen.pdf.

## Culprit

Not yet identified. `4d06e5f3` ("perf(image): resolve the DeviceN tint transform per sample, not per pixel", #282) was the obvious suspect and is **clean** — it renders devicen.pdf at 0.000%. A bisect over the remaining 83 commits is in progress and will be reported separately.

## Affected packages

- [ ] `pdf_graphics`
- [x] Documentation / CI / repository metadata only (render artifacts)

## Change type

- [x] Test/fixture artifact update

## Validation

- [x] Full PDF.js render suite re-run on current `main` (includes #387); devicen.pdf is the only file whose output changed
- [x] `index.html` regenerated by the same full run, so its captions stay complete and consistent

## Notes for reviewers

If you would rather not record the regression, the alternative is to leave the gallery stale and track the drift purely in the issue tracker — say the word and I will close this. Either way the underlying rendering change still needs a fix.


ben-milanko:
## Culprit found: `8092e7d2`, and it exposed a pre-existing bug

Bisect result, verified directly against the commit and its parent:

```
302b16fa (8092e7d2^)  GOOD  0.000%
8092e7d2              BAD   3.785%
```

`8092e7d2` is "refactor(pdf_graphics): extract a shared PdfColorSpace module (#310) (#341)".

### What actually broke

The green and blue triangles render **black**; the red one is still correct.

`devicen.pdf` paints through `[/DeviceN [/Red /Green /Blue] [/ICCBased …] 14 0 R]`, where object 14 is a **3-input** `FunctionType 0` sampled tint transform (`/Domain [0 1 0 1 0 1]`, `/Size [13 13 13]`, 3 outputs) whose sample data is essentially identity.

Before the refactor, `mesh.dart` called `colorFromComponents(values, components)`, which **bypassed the tint transform entirely** and treated the three DeviceN components as RGB. Because the colorants happen to be named Red/Green/Blue and the transform is near-identity, that was *accidentally* correct for this file.

The refactor routed vertex colours through `PdfColorSpace.toSrgb`, which correctly applies the tint transform — and that surfaced the real defect:

**`_SampledFunction` in `packages/pdf_graphics/lib/src/function.dart` is strictly one-dimensional.**

- `function.dart:94` keeps only `size[0]`, discarding the remaining `/Size` entries.
- It never overrides `evaluateAt`, so the base implementation at `function.dart:18` (`evaluate(inputs[0])`) **discards every input but the first**.

Evaluating the file's own tint transform confirms it:

```
[1.0, 0.0, 0.0] -> [1.000, 0.000, 0.000]   correct (red)
[0.0, 1.0, 0.0] -> [0.000, 0.000, 0.000]   should be green
[0.0, 0.0, 1.0] -> [0.000, 0.000, 0.000]   should be blue
[1.0, 1.0, 0.0] -> [1.000, 0.000, 0.000]   should be yellow
```

Only the first input reaches the sampler, so every colour whose first component is 0 collapses to black — exactly the observed render. Type 4 (PostScript) functions already override `evaluateAt` and handle multiple inputs; type 0 does not.

### Implication for this PR

The refactor is *more* correct, not less — the regression is that it removed an accidental bypass that was masking a missing feature. So this is a genuine bug with a clear fix (multi-dimensional sampled functions with multilinear interpolation), not something to bake into the gallery.

**Recommend holding or closing this PR** and fixing `_SampledFunction` instead, which should restore devicen.pdf to 0.000% on its own merits.

PR #429, opened 2026-07-20
fix(render): paint sub-pixel strokes as solid hairlines, not faded lines (#426)

### LINKED ISSUE #426: render: hairline strokes have no device-pixel floor, so thin linework renders pale
## Symptom

On dense CAD/schematic drawings our strokes render **lighter and fuzzier** than other PDF viewers: hairlines come out as pale grey rather than solid black, and thin geometry loses contrast. Reported from the field on Windows (2.0.0), with a side-by-side against another viewer showing our lines visibly washed out at the same zoom.

## Root cause: no device-pixel floor on stroke width

The interpreter converts stroke width from user space to page space (`interpreter.dart:1509-1512`):

```dart
width: _state.stroke.width <= 0
    ? k                       // zero width = thinnest line
    : _state.stroke.width * k,
```

and the device turns that straight into a Flutter paint with **no minimum** (`canvas_device.dart:360`, `:394`):

```dart
..strokeWidth = stroke.width
```

The canvas is scaled by `pixelRatio` before the device paints (`renderer.dart:522`, `:554`), so:

```
device pixels = page units x pixelRatio
```

Two distinct defects fall out.

### 1. Sub-pixel widths have no floor

PDF §8.4.3.2 leaves thin strokes to the device, and every mainstream renderer clamps the painted width to **at least one device pixel** so a hairline stays a solid line. We do not, so a sub-pixel width is handed to Skia and anti-aliased to partial coverage — a 0.06 pt line at `pixelRatio` 1 covers **6% of a pixel** and paints as ~6%-alpha grey.

This is not a corner case. The reported drawing (`VT-DRG-KAV-GGE-SIG-0003`, 2 pages, 106,764 content ops) sets these widths:

| `w` operand | occurrences |
|---|---|
| **0.0** | 88 |
| **0.12** | 77 |
| **0.06** | 32 |
| 0.24 | 10 |
| 0.66 | 2 |
| 0.18 | 2 |

So ~109 of 211 width settings are sub-pixel at ordinary zoom, and they are the majority of the linework.

### 2. Zero width is resolved in the wrong space

`w 0` maps to `k` — one **user-space** unit — not one **device** pixel. At `pixelRatio` 1 that lands on ~1 px by coincidence; at `pixelRatio` 3 (a deep-zoom raster) the same "thinnest possible line" paints **3 px wide**. So zero-width strokes are simultaneously the wrong width and inconsistent across zoom levels, while the sub-pixel widths above are too light.

## Proposed fix

Clamp in **device space**, where the answer is well defined:

- Give the painting device the output scale it is being used at (the renderer already knows `pixelRatio` at every call site that builds a `CanvasPdfDevice`).
- Paint with `max(width, 1 / pixelRatio)` in page units, i.e. never less than one device pixel.
- Resolve `w 0` through the same floor rather than to `k`, so "thinnest line" means one device pixel at every zoom.

The same treatment is needed in the strip rasterizer, which reads `stroke.width * sf` directly (`raster/strip_generator.dart:268`, `:499`, `raster/strip_binning_device.dart:234`).

## Validation this will need

This changes the pixels of **every stroked page**, so it cannot land without:

- Ghent baseline review (`GHENT_UPDATE`) — expect widespread, intentional diffs.
- The PDF.js pixel comparison, which is the useful signal here: if the clamp is right, agreement with the reference renders should **improve**, not just change.
- A visual gallery pass on the CAD corpus.

Worth doing as its own PR with the baselines refreshed deliberately, not folded into anything else.

## Not this ticket

The same report also shows **fuzzy glyphs**, which is a different cause — an upscaled resolution-capped raster rather than stroke geometry. That belongs with #405 / #374 (the per-page raster cap and deep-zoom discard rate); the memory profile from the same session shows a 4.59 GB RSS high-water and a 55% tile discard rate on this document.

ben-milanko:
**Draft — do not merge yet.** The fix is right and measured, but it changes pixels on every stroked page and the baseline work below has not been done.

## The cause is alpha, not width

I filed #426 assuming thin lines were rendering *thinner*. They are not. `strip_binning_device.dart:228` already documented the real mechanism:

> Skia renders strokes thinner than one device pixel as a 1-px hairline **with the alpha modulated by the width** (not a true sub-pixel band)

So a 0.06 pt line is painted one pixel wide at **~6% opacity**. That is exactly the reported symptom — correct thickness, washed out — and it is most of the drawing: the reported file sets width **0.06 or 0.12 on 109 of its 211 `w` operations**.

## Changes

- **`CanvasPdfDevice`** floors stroke width at one device pixel by emitting Skia's **hairline** (`strokeWidth = 0`), which the SDK documents as one device pixel at any transform, at full alpha. A hairline rather than a page-space floor because one recorded picture is replayed at several ratios (base raster, deep-zoom detail patch, retained scene) — a floor baked at record time is right for one and too thick for the rest.
- **`strip_binning_device`** stops scaling alpha by the sub-pixel width, so both raster paths agree. It was faithful to Skia, but not to the page.
- **The interpreter** stops resolving `w 0` to one *user*-space unit. §8.4.3.2 defines it as one *device* pixel, so it is a device decision; in user space it was correct only at ratio 1 and three pixels wide at ratio 3.

## Performance: 1.56x faster, against expectation

Reported drawing, 2 pages at scale 2, min of 3 interleaved reps:

| | renderMs |
|---|---|
| before | 409 |
| after | **262** |

I predicted this would cost fill time (0.06 px → 1 px is ~16x the covered pixels). Wrong: Skia's hairline rasterizer skips stroke-to-fill conversion entirely — no geometry expansion, no join/cap tessellation — and on a thin-stroke-heavy page that geometry work dominates the fill. Every statistic favours the change; the new path's *worst* run beat the old path's worst.

## What still has to happen before this can land

- **Ghent baselines**: 15 render tests fail, expected and intentional. They need refreshing with per-file review, not a blanket `GHENT_UPDATE`.
- **PDF.js pixel comparison**: this is the real signal. If the change is correct, agreement with the reference renderer should **improve**. Unverified so far — and it is the check that would tell us the hairline is right rather than merely different.
- **`canvas_line_replay_test`** now fails by **one alpha unit at one pixel**. It asserts the cached-Paint path is pixel-identical to fresh Paints; a 0.5-width stroke now becomes a hairline, and Skia appears to batch hairlines differently between the two. Small, but it is a real interaction and should be understood rather than re-baselined.
- `editing_thumbnail_view_test` also failed in the same run; not yet triaged, possibly unrelated.

Everything else passes: `strip_parity_test` (the canvas/strip invariant) is green, `pdf_graphics` 870 pass.

Closes #426 once the above is done.

https://claude.ai/code/session_01KHF8boHzZGxFNijgPRSuaX


ben-milanko:
Both loose ends cleared; taking this out of draft.

**Baselines done.** The decisive check is the PDF.js pixel comparison, and it is unambiguous:

| | pass | fail |
|---|---|---|
| main | 64 | 100 |
| this branch | **68** | 96 |

Of the **30 pages whose render changed, 29 moved closer to the reference and 0 moved away**. `standard_fonts` went from ~10–12% differing pixels to ~3–5%; `radial_gradients` from ~3–4.8% to ~0.6–1.1%. That was the criterion for "correct rather than merely different".

**Ghent: 13 baselines updated, not 32.** `GHENT_UPDATE=1` rewrites every drifted baseline, including 18 the suite already tolerates as *"known baseline deviation — colour handling is not pixel-exact"*. Blanket-accepting would have folded unrelated pre-existing drift into this PR and buried the fact that **DeviceN images do not render at all** (now #430, with #431 tracking the wider set of self-grading Ghent pages that render their own failure marker). `GWG030` is also excluded — it is the standing pre-existing failure on main and its diff is a colour change, not a stroke one. Ghent is now back to main's exact state: 39 pass / 14 deviation / 1 fail.

**`canvas_line_replay_test`** — my "Skia batches hairlines differently" theory was **wrong**. A probe rendering each width in isolation, and the full sequence, found zero differing bytes. The real cause: the test's reference is a hand-rolled `canvas.drawPath` using the raw width, bypassing the device, so it was asserting pre-floor behaviour. The floor is device policy rather than part of the fast path being guarded, so the reference now applies it via the device's own helper (one source of truth).

**`editing_thumbnail_view_test`** — passes standalone (12/12); flaky under full-suite load, unrelated.

Full editor suite: **1780 pass, only the pre-existing ghent GWG030** — same as main.

Issue #427, opened 2026-07-20
perf(editor): 4.6 GB RSS and 55% tile discard on a 559 KB 2-page CAD drawing


ben-milanko:
## Report

Windows, 2.0.0 release, dense signalling drawing (`VT-DRG-KAV-GGE-SIG-0003`, **2 pages, 559 KB**, 106,764 content ops, only two 145x385 grey JPEGs). User reports serious slowdowns.

The devtools export shows this is **not** a frame-time problem:

```
frames:  fps 50.4   avgBuild 0.68 ms   avgRaster 1.66 ms   worst 6.0 ms   jankFrames 0
```

It is memory:

```
rssBytes           2.02 GB
rssHighWaterBytes  4.59 GB          <- from a 559 KB, 2-page file
cacheTotalBytes    211 MB  (ceiling 402 MB)
  decoded-image    3 entries / 179 MB     (~60 MB each, ~15 Mpx)
  tiles           30 entries / 31.5 MB
tileStore:  scheduled 8495   landed 3799   discarded 4696   batches 1249
deepZoomMode: tiles (batched)   renderWorkers: 3
```

A 4.59 GB high-water on Windows means paging, which is what the user feels. The document has no large images, so those three ~15 Mpx cache entries are **page rasters**, not document images.

## What this is evidence for

Two existing tickets, with real numbers attached for the first time:

- **#405 (global live-raster memory budget).** The per-page 16.7 Mpx cap is applied independently per page view and again for the detail patch, with no cross-page accounting. Three live rasters at ~60 MB is exactly that shape, and it is what pushes RSS into GB territory on a file that is 559 KB on disk.
- **#374 (deep-zoom discard rate).** **4696 of 8495 scheduled tiles were discarded — 55%.** More than half the tile work on this session was thrown away, and each discarded tile still cost a raster.

## Not addressed by anything merged since the release build

Worth stating explicitly, because the obvious guess is wrong: **#383's grid-cull region replay does not engage here.** It escalates only *above* `spatialRegionReplayMaxCommands = 250000` (`retained_scene.dart:112`); at 106,764 ops this page stays on the unchanged linear path. #388's banding engine is not wired into the page view yet, and #423 only helps text extraction (31 ms in this session — irrelevant).

So nothing merged after the 2.0.0 build changes this document's behaviour.

## Notes

- The file is in the local corpus as `corpus/sig-arrangements.pdf`.
[...]

Issue #405, opened 2026-07-19
perf(editor): global live-raster memory budget


ben-milanko:
## Summary

Per-page raster caps are additive with no global budget. Each live page may simultaneously hold a base raster up to 16.7 Mpx (~64 MB RGBA), a detail patch given the same budget, a retained scene (~31 MB cited for one dense sheet), a preview clone, and (web) a Slug picture. With 3-5 pages live in the cacheExtent, that is several hundred MB of GPU memory with no cross-page accounting.

## Evidence

- `packages/dart_pdf_editor/lib/src/pdf_page_view.dart:408-419` — `_maxPixels = 1<<24` per page (line 411).
- `pdf_page_view.dart:1806-1814` — `_detailGeometryAt` clamps the detail region by the same `_maxPixels`.
- Retained-scene memory cited at `pdf_page_view.dart:218-224`.
- The preview/full-raster caches are budgeted (`preview_cache.dart:33-58`, 32 MiB default) and memory-pressure handling exists (`pdf_viewer.dart:1454-1460`), but live-page rasters are exempt.
- `_maxPixels` is ~4x what any current display can show (4K = 8.3 Mpx).

## Impact

Memory pressure / jetsam on iOS when reading large-format documents zoomed; GPU eviction churn.

## Proposed fix

1. Introduce a global live-raster pixel budget: register each `_PdfPageViewState`'s base+detail pixels, evict/downgrade farthest-from-viewport pages first.
2. Lower `_maxPixels` toward viewport^2 x maxUsefulZoom.
3. Count retained-scene decoded images against the same budget.

## Dependencies

None.

## Suggested order

Phase 3, ticket 14 of 16.

ben-milanko:
## Verification pass — problem CONFIRMED, but the scope is much smaller than written

**The defect is real.** `pdf_page_view.dart:433-436`: `static const _maxPixels = 1 << 24;` (~16.7 Mpx) and `_maxDimension = 8192.0`, both per-`_PdfPageViewState`, applied independently in `_effectiveRatioAt` (`:780`) and **again** for the detail patch (`:1907` — the comment there says "the patch obeys the same pixel budget as the base", i.e. additive by design). No summation across live pages, patches or `PdfRetainedScene`s.

**Two pieces of infrastructure already exist or are in flight, and this ticket should be rewritten to sit on them rather than duplicate them.**

[...]

Issue #374, opened 2026-07-18
perf: tile pyramid — mobile budget scaling, iOS validation, deep-zoom discard rate


ben-milanko:
Follow-up to #314 (tile pyramid shipped + on by default on all platforms). These are the residuals from the on-device validation pass that didn't block the core feature but should be tracked.

## 1. Mobile tile budget scaling
`PdfTileStore` uses the same **96 MB** budget on mobile as desktop (`_defaultMaxBytes`, no platform scaling). On an Android release trace opening large CAD books, `rssHighWaterBytes` peaked at **3.57 GB** (that peak was progressive-open file buffers + big-book rasterization, *not* tiles — tiles were 0 bytes at the time). A deep-zoom session then pushed tiles to their full 96 MB. On a lower-RAM device, 96 MB of pyramid stacked on an already-high baseline is a plausible OOM/jetsam risk.

- Proposal: a `pdfDefaultTileBudgetBytes()` in `performance_policy.dart` (mirroring the image-cache budget policy) returning ~48–64 MB on mobile, wired into `PdfTileStore.instance`. The budget-vs-demand guard already falls back to the single patch when a view won't fit, so a smaller mobile budget just trips that fallback a bit sooner.
- `PdfTileStore.maxBytes` is already live-adjustable, so this is a policy wiring change.

## 2. iOS tile validation
Deep-zoom tile behavior was confirmed on-device only on **Android** (no thrash, jank 10/120, settled under budget). **iOS** (CanvasKit-adjacent `toImageSync`/Impeller behavior) has not had a device deep-zoom pass. Verify: green-grid stability, `scheduled`/`batches` flat on a static view, no seams over the Slug layer.

## 3. Deep-zoom discard rate on mobile
The Android deep-zoom trace showed `discarded 170 / scheduled 436` (~40%) — tiles rastered then thrown away as **stale** (page invalidated after dispatch), distinct from eviction. Likely transient: deep-zooming while the scene is still upgrading vector-first → full image-bearing record, each swap invalidating in-flight tiles (longer settle window on mobile). Confirm it's the settle window and not something re-invalidating mid-zoom; ~40% wasted raster is CPU/battery on mobile.

## Context
- Tile pyramid: #314 (closed). Budget-vs-demand guard: #371. Default-on flip: `b160061d`.
- Dev-log: `doc/dev-log/2026-07-18-tile-budget-thrash.md`.

ben-milanko:
[...]

PR #433, opened 2026-07-20
fix(editor): scope tile invalidation to the pages it names (#374)

ben-milanko:
## What

`PdfTileStore.invalidate(pages:)` documents that only the named page slots' in-flight rasters are discarded. It didn't behave that way: `_isStale` also tested a shared `_globalEpoch` that *every* invalidation bumped, so any targeted invalidation threw away every other page's in-flight tiles.

## Why it matters

This is not a rare edge case — it fires throughout ordinary deep-zoom reading.

`_PdfPageViewState._updateDetail` calls `_dropDetail()` whenever `_detailGeometryAt` returns null, and that includes **any page scrolled off screen** (`pdf_page_view.dart:1894` returns null on an empty visible rect). `_dropDetail()` invalidates that page's tile slot.

So on a two-page document, every settle of the **off-screen** page discarded the tile batch in flight for the page the user was actually panning. The more you pan, the more the other page starves you.

This is the concrete mechanism behind the discard rate reported in #427: **4,696 of 8,495 scheduled tiles (55%) discarded** in one session on a 2-page drawing. Each discarded tile still paid for its raster.

## The fix

Split the ordering counter from the staleness test:

- `_epochCounter` — monotonic, stamps dispatches and invalidations in order. Deliberately *not* the staleness test.
- `_allPagesEpoch` — raised only by a full `invalidate()`.
- `_pageEpoch` — per page slot, as before.

A targeted invalidation can no longer reach across pages. `invalidate()` with no argument is unchanged.

## Tests

Two new tests, **both fail on main and pass here**:

- `keeps in-flight rasters for the pages it did not invalidate` — two pages with 4 tiles each in flight, `invalidate(pages: {0})`. Main discards **8**; expected and now actual is **4**.
- `an off-screen neighbour settling does not starve the panned page` — replays #427's session shape (8 pan settles on page 0 interleaved with page 1's off-screen invalidations). Main discards tiles on every settle; now **0 discarded, all scheduled tiles land**.

Full `dart_pdf_editor` suite: 1775 pass, only the pre-existing `ghent_render_test` GWG030 failure, identical to main. `dart analyze` clean.

## Scope

This addresses the discard half of #427. The memory half (#405's global live-raster budget) is untouched and still open — see the note I'll add to #427.

Refs #374, #427.

### DIFF
--- a/packages/dart_pdf_editor/lib/src/tile_store.dart
+++ b/packages/dart_pdf_editor/lib/src/tile_store.dart
@@ -279,9 +279,17 @@ class PdfTileStore extends ChangeNotifier {
 
   /// Invalidation epochs: a tile whose page was invalidated after its raster
   /// was dispatched is discarded on completion (mirrors the render worker's
-  /// stale-inflight guard). [_globalEpoch] rises on every invalidation;
-  /// [_pageEpoch] records the epoch at which each page was last invalidated.
-  int _globalEpoch = 0;
+  /// stale-inflight guard).
+  ///
+  /// [_epochCounter] only stamps dispatches and invalidations in order - it is
+  /// deliberately NOT the staleness test. Only [_allPagesEpoch] (raised by a
+  /// full [invalidate]) and [_pageEpoch] (per page slot) make a tile stale, so
+  /// a targeted invalidation cannot reach across pages. That distinction is
+  /// load-bearing: the page view invalidates its own slot on every live
+  /// transform tick, and a shared counter would make each of those ticks throw
+  /// away every other live page's in-flight rasters (issue #374).
+  int _epochCounter = 0;
+  int _allPagesEpoch = -1;
   final Map<int, int> _pageEpoch = {};
 
   bool _tickScheduled = false;
@@ -595,7 +603,7 @@ class PdfTileStore extends ChangeNotifier {
       pending.add(key);
       return;
     }
-    final dispatchEpoch = _globalEpoch;
+    final dispatchEpoch = _epochCounter;
     rasterize(region, ratio).then((image) {
       _inFlight.remove(key);
       if (_disposed || _isStale(id.pageIndex, dispatchEpoch)) {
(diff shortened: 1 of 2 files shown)


ben-milanko:
## Review findings — both addressed in 5c6b50f7

### 1. The new `_allPagesEpoch` arm of `_isStale` had zero coverage

Splitting the shared counter created a coverage hole. `null clears every page` calls `raster.flush()` **before** `invalidate()`, so it only ever exercises cache clearing — never the in-flight staleness guard. Before this PR that arm was covered *incidentally*, because the targeted tests drove the same shared counter; the split silently removed that.

Proven by mutation: replacing `_allPagesEpoch > dispatchEpoch` with `false` — deleting the entire full-invalidation guard — **passed all 27 tests** in the package. `invalidate()` is also the memory-pressure path, so a regression there would leak stale tiles under exactly the conditions the store exists to handle.

Added `null discards in-flight rasters on every page`: two pages, 8 rasters in flight, full invalidation. Kills the mutation.

### 2. Both new comments misattributed the trigger

They justified the split with *"the page view invalidates its own slot on every live transform tick"*. That path is `_onLiveTransformChanged`, which is gated on `_slugPicture != null` → `_slugPictureScene` → `kIsWeb`. It is **web-only**, so it is not what makes this bug matter on the reported Windows session.

Corrected to the path actually verified: `_updateDetail` calls `_dropDetail()` whenever `_detailGeometryAt` returns null, which includes every page scrolled off screen (`pdf_page_view.dart:1894`). The PR description had this right; the code comments did not, and the comment is where the next reader looks.

### Checked and found clean

- **Epoch ordering.** Walked dispatch-before/after against targeted and full invalidations, including the boundary where a dispatch and an invalidation share a stamp. Bumping the counter on targeted invalidations is *required* — without it a dispatch stamped at the same value would compare not-stale under `>` despite happening first.
- **`_pageEpoch` growth.** Bounded by page count, and after a full invalidation every entry is dominated by `_allPagesEpoch`. Dead weight, not a leak — left alone rather than churn the diff.
- **`_allPagesEpoch = -1` initial.** Matches the `?? -1` default on the `_pageEpoch` lookup; a pre-invalidation dispatch stamped 0 is correctly not stale.
- **Re-schedule window.** A tile invalidated while in flight stays in `_inFlight` until its stale raster completes, so `_schedule` refuses to re-queue that key until then. Pre-existing behaviour, unchanged here.

### Verification

Full `dart_pdf_editor` suite: **1777 pass**, only the pre-existing `ghent_render_test` GWG030 failure — identical to main. `dart analyze` clean. All three new tests fail on main.

Issue #404, opened 2026-07-19
perf(editor): cache per-annotation appearance pictures


ben-milanko:
## Summary

In editing mode, `_AnnotationAppearanceLayer` re-renders **every** annotation appearance on a page after each revision — 50 ink strokes/highlights means 50 appearance renders after every single stroke commit, though only one annotation changed. Each appearance render does its own ImageCollector scan + decodeImages + interpreter walk.

## Evidence

- `packages/dart_pdf_editor/lib/src/pdf_viewer.dart:5503-5589` — collect at 5578-5584, `_render` at 5572.
- Per-annotation cost: `packages/dart_pdf_editor/lib/src/renderer.dart:352-377` (`renderAnnotationPicture`: scan + decode + walk each).

## Impact

Editing latency after every commit on annotated pages; redundant walks and image decodes on markup-heavy pages.

## Proposed fix

Cache per-annotation pictures keyed on (annotation object identity, appearance stream identity); re-render only the edited annotation(s) — the controller knows the touched `/Annots` slots.

## Dependencies

- Blocked by #393 (parsed-ops cache) — the appearance walk should hit the ops cache; the picture cache layers on top.

## Suggested order

Phase 3, ticket 13 of 16.

ben-milanko:
## Verification pass — CONFIRMED, and the root cause is #397

`_AnnotationAppearanceLayer` is at `pdf_viewer.dart:5618`. `_render()` (`:5693`) fans out `Future.wait([for (final annotation in annotations) _renderAnnotation(...)])` over **all** non-hidden annotations with an appearance. `_pictures` is a plain `List<ui.Picture>` rebuilt wholesale at `:5721` — no cache keyed on appearance-stream identity anywhere in the state. `keepCurrent` (`:5657`) only avoids *clearing* the old pictures; it does not skip the re-render.

**Why it fires every revision — worth adding to the ticket:** the trigger is `didUpdateWidget` at `:5651`, `if (!identical(oldWidget.page, widget.page) || ...)`. `_pages` is rebuilt from scratch on every revision (`pdf_viewer.dart:2270`, `_pages = [for (var i = 0; i < count; i++) document.page(i)];`) and `PdfDocument.page()` returns a fresh `PdfPage` each call (`pdf_document/document.dart:157`). So page identity **always** differs, and the identity guard never holds.

That means there are two independent fixes and they should be sequenced:
[...]

Issue #422, opened 2026-07-20
ci: web-worker auto-regen races, flip-flops, and leaves PR heads without checks


ben-milanko:
## Summary

The CI auto-regeneration of the bundled web render worker (#411 → #412) has three defects that showed up together on #417. Net effect: **three bot commits pushed to one PR branch, the bundle oscillating between two different outputs, and a head commit with no CI checks at all**, so the PR could not go green until it was manually closed and reopened.

None of this is wrong about the *goal* — the asset must stay checked in (it is a declared Flutter asset that ships to pub.dev) and a human should not have to remember to rebuild it. It is the push-back mechanism that needs hardening.

## Evidence (from #417)

Three consecutive commits by `github-actions[bot]`, all titled *"chore(editor): rebuild the bundled web render worker"*:

| commit | bundle blob | bytes | committed |
|---|---|---|---|
| `4ebff786` | `6a3b87e7…` | 976190 | 03:07:13 |
| `31077941` | `87447f9c…` | **976193** | 03:07:18 |
| `287e6fb3` | `6a3b87e7…` | 976190 | 03:26:36 |

and on the PR:

```
$ gh pr checks 417
no checks reported on the 'perf/page-and-annotation-cache' branch
```

A local rebuild on the pinned SDK (3.44.4) matched `6a3b87e7…` / 976190 byte-for-byte, so the *content* at head was correct — the problem is entirely in how it got there.

## Defect 1 — superseded runs still push (no concurrency group)

`.github/workflows/ci.yml` declares no `concurrency:` group. Two pushes in quick succession leave two `worker-bundle` jobs running at once; both rebuild, both push. Hence two bot commits **five seconds apart**.

**Fix:** add a cancel-in-progress concurrency group, e.g.

```yaml
concurrency:
  group: ci-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

## Defect 2 — the reproducibility check is weaker than it looks

`tool/ci/regen_web_worker.sh` builds twice and requires the two to byte-match:

```
# Reproducibility: a second build must byte-match the first, confirming the
# output is not machine-dependent ...
```

That establishes determinism **within one runner in one run**. It does not establish determinism **across** runs — and the table above shows the output genuinely differs between runs (976190 vs 976193) on the same source. So even with defect 1 fixed, successive runs can disagree and each push "its" bundle, producing an A→B→A ping-pong.

[...]

PR #441, opened 2026-07-20
perf(document): cache PdfPage instances, resolving inheritance live (#418)

### LINKED ISSUE #418: perf(document): cache PdfPage instances so the annotation cache is reachable
## Problem

`PdfDocument.page(int)` builds a **fresh `PdfPage` on every call** (`document.dart:157`). `PdfPage.annotations` parses `/Annots` lazily and caches it *per instance* (`page.dart:57-68`), so that cache is effectively unreachable: every access re-resolves the array and rebuilds every `PdfAnnotation.fromDict`.

Viewer code reads annotations per page per frame, so a page with N annotations re-parses all N repeatedly.

There is a second consequence, noted on #404: `_AnnotationAppearanceLayer` guards re-rendering with `identical(oldWidget.page, widget.page)` (`pdf_viewer.dart:5651`), and that check can *never* hold, because `_pages` is rebuilt from `document.page(i)` every revision (`pdf_viewer.dart:2270`). Caching pages would make that guard meaningful.

## Why the obvious fix does not work

I prototyped `Map<int, PdfPage> _pageCache` on `PdfDocument` while doing #397 and backed it out. It is **not sound today**:

`PdfPage` snapshots its **merged inherited attributes** — `/Rotate`, `/Resources`, `/MediaBox`, `/CropBox` — at construction (`document.dart:177-183` passes `merged.*` into the constructor; `page.dart:19-29` stores them). `PdfEditor` mutates page dictionaries **in place**. So a cached page keeps reporting pre-edit values.

Concretely it fails `form_fill_test` — *"form appearances stay upright when document rotation is baked in"*:

```dart
// editor.dart rotatePage
final page = document.page(index);          // caches the page
page.dict['Rotate'] = CosInteger(next);     // mutates the dict in place
_updater.markChanged(page.dict);
_regenerateFormAppearancesOnPage(index);    // re-reads the CACHED page -> old rotation
_markContent([index]);                      // invalidation would happen here: too late
```

So the natural hook — invalidating from `PdfEditor`'s `_mark*` impact methods — **runs after** the code that already consumed the stale page. I tried exactly that; the test still fails.

The real choke point is `markChanged` (every in-place dictionary mutation), which has **44 call sites** across `pdf_document` (`struct_tree_editor`, `signature_editor`, `page_editor`, `content_editor`, `form_admin`, …).

## Two viable approaches

1. **Make `PdfPage` prefer its live dictionary over the snapshot.** Store the *inherited-only* values and resolve on access as `dict['Rotate'] ?? _inheritedRotate` (same for `/Resources`, `/MediaBox`, `/CropBox`). Inherited values only change on structural edits, which already call `invalidatePageCache()`. This fixes the staleness at the source and makes instance caching safe. Requires changing `_findPage` to stop pre-merging and `PdfPage` to merge lazily.

2. **Route every mutation through an invalidating wrapper.** Add `PdfEditor._markChanged(dict)` that calls `_updater.markChanged(dict)` *and* drops cached page objects, then convert all 44 call sites. Mechanical but broad, and it only holds as long as nobody calls `_updater.markChanged` directly again.

(1) is better: it removes the hazard rather than papering over it, and it is contained to two files.

## Interim state

#417 landed the other half of #397 (O(1) `pageIndexOf`), which needed no caching and is independent of this.

A narrower `invalidatePageObjects()` (clearing page objects without discarding the leaf walk) was part of the prototype and is worth keeping in mind for whichever approach is taken — dropping the whole page-tree cache per mutation costs more than the cache saves.

## Related

- #397 (parent; lookup half done in #417)
- #404 — the appearance layer's `identical` guard that this would make functional
- #414 — the same family of problem: identity used as a change signal

ben-milanko:
## What

`PdfDocument.page()` built a fresh `PdfPage` per call, so `PdfPage.annotations`' lazy cache was **unreachable** — every access re-resolved `/Annots` and rebuilt every `PdfAnnotation`. The viewer reads annotations per page per frame.

## Why the obvious fix was backed out before

`PdfPage` snapshotted its **merged** inheritable attributes at construction, while `PdfEditor` mutates page dictionaries **in place**:

```dart
final page = document.page(index);        // would cache the page
page.dict['Rotate'] = CosInteger(next);   // mutates the dict in place
_regenerateFormAppearancesOnPage(index);  // re-reads the page in the SAME call
```

No invalidation hook can run early enough, which is why the ticket records an earlier prototype being reverted.

## Approach

Option 1 from the ticket — fix the staleness at its source rather than route 44 `markChanged` call sites through a wrapper.

`PdfPage` now holds only what its **ancestors** supply and reads its own `/Resources`, `/MediaBox`, `/CropBox`, and `/Rotate` from `dict` on access. Inherited values change only on structural edits, which already call `invalidatePageCache()`. With the snapshot gone, instance caching is safe; `_pageCache` clears alongside the leaf index.

## The hazard caching introduces, and how it is bounded

A cached page's annotation list now outlives edits, so it has to notice them. It is keyed on the **identity of the resolved `/Annots` array plus its length**, covering how the editors actually mutate it — replacing the array, and adding or removing entries in place.

Editing an existing annotation's properties needs no invalidation (`PdfAnnotation` reads through to its own dictionary, mutated in place), so a same-array, same-length change is correctly a hit. A wholesale item *substitution* preserving both identity and length would be missed; nothing does that today, and the doc comment says so rather than leaving it implicit.

## Measured

20 passes over 10 pages carrying 60 annotations:

| before | after | |
|---|---|---|
| 0.45 ms | 0.03 ms | **15×** |

The absolute figure scales with annotation count; the ratio is the point, since the viewer repeats this per frame.

## Tests

Six new tests; **four fail on main**, including:

- a cached page sees an in-place `/Rotate` edit — the exact case that sank the earlier attempt
- a cached page sees an in-place `/MediaBox` edit
- the annotation cache notices entries appended straight into the live array
- `page(i)` hands back one instance

Plus a guard that a page with no entry of its own still inherits from its ancestors, since that is what the constructor change could plausibly break.

## Verification

- `pdf_document`: **762 pass** — including `form_fill_test`'s *"form appearances stay upright when document rotation is baked in"*, which the ticket names as the blocker
- `dart_pdf_editor`: **1782 pass**, only the pre-existing ghent GWG030
- `dart analyze` clean

## Follow-on

This also makes `_AnnotationAppearanceLayer`'s `identical(oldWidget.page, widget.page)` guard meaningful (`pdf_viewer.dart:5651`) — it could never hold before. Realising that is #404's business, not this PR's.

Closes #418.


ben-milanko:
Rebased onto current main (which has since taken #438–#446) and re-validated.

Note this PR now carries **both** #418 and #404 — #443 was merged into this branch rather than main, so the page-instance cache and the annotation appearance cache land together.

**The rebase:** the only conflict was the generated `pdf_render_worker.dart.js`. The bot's earlier rebuild was generated from a superseded source state, so it was **skipped** and the bundle regenerated from the rebased tree — two dart2js outputs cannot be reconciled hunk by hunk (#422).

**Re-validated on the rebased tree, not just before it:**
- `pdf_document`: **770 pass**
- `dart_pdf_editor`: **1786 pass**, only the pre-existing ghent GWG030
- `dart analyze` clean

`mergeable` is back to MERGEABLE.

Issue #451, opened 2026-07-21
perf(web): the visible page's own record ships full-resolution RGBA (progressive first-record resolution)


ben-milanko:
> **Rescoped 2026-07-26.** Sub-problem (2), *prefetch starving the visible page*, is done — the visible page now preempts prefetch (#456), prefetch records ship capped image resolution (#473), and the CMYK-conversion cost that made prefetches long enough to collide is halved (#598). What remains is sub-problem (1): **the focused/visible page's own record is still uncapped**, so a Flate-coded page still ships full-resolution RGBA at 2× headroom. The original two-part ticket is preserved in the history block below.

## What's left

The **visible page's first record** still ships full-resolution RGBA at `imageCap=2.00`. #473 capped image resolution for *prefetched / off-focus* pages only and explicitly deferred the visible page by name: *"the visible-page record itself (needs progressive first-record resolution, not a prefetch cap)."* No merged PR has delivered that.

For image-dominated (scanned / CAD / Flate-artwork) documents this is the whole problem — a single visible page's record still reaches ~40–60 MB, and the glyph-outline dedup that helped text-heavy docs (#601, −31%) contributes nothing there.

### Acceptance

The first image-bearing record for the **focused** page ships at a reduced image resolution and refines to full resolution progressively (the progressive-first-record idea, now with the trace evidence below behind it), so a Flate-coded page like the 39 MB page 1 lands its first record at a fraction of that. Prefetch/off-focus capping (#473) stays as is.

### Loose end surfaced by #601

`PdfPageRenderScheduler`'s "one page-walk per frame" serialization has silently lapsed: worker-path builds run in record-arrival order rather than focus order, and every focus page is granted twice (the "record fetched twice" observation in the history below is the same defect). Confirm whether this belongs here or in its own ticket.

## Already landed (prefetch half + related)

- **#456** — the visible page leaves its sticky worker when that worker is busier, so it no longer waits behind a background prerender.
- **#473** — caps image resolution for prefetched/off-focus records (page-1 neighbours 45 → 11.8 MB); frees the single effective worker.
- **#598** — memoizes colour conversion, halving image-decode time (DeviceCMYK −66%), removing the serialize anomaly that made prefetches collide with the visible page.
- **#601** — dedups glyph outlines per record (record format v3→v4); big win on font/text-heavy docs, no effect on image-dominated ones.

---

<details>
<summary>Original ticket (2026-07-24) — kept for history; the prefetch half is now resolved</summary>

[...]

PR #456, opened 2026-07-21
perf(editor): let a page leave its sticky worker when that worker is busier

ben-milanko:
## The measurement

Device trace via #455's phase split — 62-page book, two workers:

```
phase worker=0 page=2 queue=1024.3ms worker=2613.0ms serialize=1817.0ms outcome=ok
phase worker=0 page=0 queue=2429.5ms worker=16.0ms   transcript=hit
raster base-full page=0 ms=80.5      <- image finally visible
```

**The visible page's image record waited 2.43 seconds for a worker that then did the job in 16 ms**, while worker 1 sat idle for most of that window. It is the single largest term in the ~4.7 s to first image — larger than worker startup, picture build, and raster combined.

## Root cause

`_workerForPage` pinned the page:

```dart
final existing = _pageWorkers[pageIndex];
if (existing != null && _workers[existing].isActive) return existing;
```

Once assigned, a page returned to the same worker for as long as it stayed active, with no regard to load. Page 0 was pinned to worker 0, which was mid-way through a 2613 ms background prerender of **page 2** — a page the user is not looking at.

## Why not simply drop stickiness

Because it is earning its keep. The sticky worker holds that page's warm transcript, which is precisely why the pinned request cost **16 ms** (`transcript=hit`). A cold worker re-records the page — that same page's first record on a cold worker took **130 ms**.

So the change abandons stickiness **only when another active worker has strictly lower load**. Equal load keeps the sticky worker: a transcript miss is never worth paying to break a tie. The trade taken here is ~130 ms against ~2430 ms.

Single-worker pools are untouched — the alternative scan skips the sticky worker itself, finds nothing, and returns it.

## Tests

Four, all **mutation-checked**:

- disabling the re-route → *"re-routes to an idle worker"* and *"a re-routed page follows its new worker"* both fail
- relaxing `<` to `<=` → *"keeps its sticky worker while loads are equal"* fails

So both the re-route and the strict-inequality boundary are pinned, not just asserted.

## Verification

`dart_pdf_editor`: **1793 pass**, only the pre-existing ghent GWG030. `dart analyze` clean.

## Scope

This is the preemption half of #451. Still open there: **page 2's serialize anomaly** — 5.3 MB taking 1817 ms while 39 MB takes 329 ms in the same worker, reproduced across three captures. That is what makes a prefetch long enough to collide in the first place, so it is worth fixing even with this routing change in.

Drafted with `opencode`/`kimi-k3`, reviewed and mutation-checked here.

Refs #451.

### DIFF
--- a/packages/dart_pdf_editor/lib/src/render_worker.dart
+++ b/packages/dart_pdf_editor/lib/src/render_worker.dart
@@ -611,7 +611,29 @@ class PdfPooledRenderWorker extends PdfRenderWorker {
 
   int _workerForPage(int pageIndex) {
     final existing = _pageWorkers[pageIndex];
-    if (existing != null && _workers[existing].isActive) return existing;
+    if (existing != null && _workers[existing].isActive) {
+      // Stickiness buys a transcript hit: the sticky worker holds this page's
+      // warm caches, so a repeat record on it is near-free, while a cold
+      // worker re-records the page. Abandon that only when the page would
+      // otherwise queue behind unrelated work - i.e. another active worker
+      // has strictly lower load. Equal load keeps the sticky worker: a
+      // transcript miss is never worth paying to break a tie.
+      var alternative = -1;
+      var alternativeLoad = 1 << 62;
+      for (var i = 0; i < _workers.length; i++) {
+        if (i == existing || !_workers[i].isActive) continue;
+        final load = _loads[i];
+        if (load < alternativeLoad) {
+          alternative = i;
+          alternativeLoad = load;
+        }
+      }
+      if (alternative >= 0 && alternativeLoad < _loads[existing]) {
+        _pageWorkers[pageIndex] = alternative;
+        return alternative;
+      }
+      return existing;
+    }
     final worker = _leastLoadedWorker(pageIndex);
     _pageWorkers[pageIndex] = worker;
     return worker;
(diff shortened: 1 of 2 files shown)


ben-milanko:
Rebased onto current main so this preview now also carries **#455's phase split**.

Ben tested the previous preview (`pr456-perf-pool-idle-routi`) and it did not feel better. That build **did** contain this change — I was wrong to say otherwise — but it was branched before #455 merged, so it reported no `wait=`/`build=`/`ms=` fields and we could not see what it did.

Reading the capture again: **this change had nothing to do in that run.** It ran with three workers and three pages, one page per worker, and the decisive request showed `queue=0.0ms`:

    [perf 2487] phase worker=0 kind=record page=0 queue=0.0ms worker=67.0ms transcript=hit
    [perf 5480] raster kind=base-full page=0     <- ~3.0 s later

Preemption cannot help a request that never queued. The collision this fixes needed **two** workers with pages 0 and 2 landing on the same one; the auto tier gave three that time.

So this PR is neither validated nor refuted by that capture — it was unexercised. It remains justified by the run where `queue=2429.5ms` behind a 2613 ms prerender, and by the fact that the pinning is unconditional regardless of load.

With the rebase, a capture from this preview will report the phase breakdown directly, so the next test says which phase actually dominates instead of leaving it to be inferred from timestamp gaps — which has misled me twice on this bug.

Issue #454, opened 2026-07-21
perf(editor): move browser JPEG decode off-thread + ship per-glyph substituted text by default


ben-milanko:
> **Rescoped 2026-07-26.** The diagnostic first step is done (#455/#457/#514), and it reframed the ticket: the **~0.9 s raster half was a mismeasurement** — instrumented cold-tab raster is ~50 ms (#457; #455 notes "my inference that raster cost 915 ms when it is 80 ms"). What is real is the **~1.2 s build window**, which splits into (a) main-thread browser JPEG decode of the page's `/DCTDecode` images and (b) command replay/shaping. Two concrete fixes remain, neither fully shipped. The original ticket is preserved in the history block below.

## What's left

1. **Move browser JPEG decode off the main thread.** #457's instrumentation attributes the largest slice of the build window to the browser codec decoding the page's `/DCTDecode` images on the UI thread (~643 ms on page 0). #588 (`decode a page's images concurrently`) overlaps that work via `Future.wait` and is default-on and pixel-identical — but it does **not** move decode off the main thread. The named follow-up (`createImageBitmap` in a worker, WebCodecs `ImageDecoder`, or decode-small-first) has not landed. This is the largest un-taken lever on the build window.

2. **Ship per-glyph substituted text by default.** #517 (`per-glyph substituted-text composition`) fixes the replay/shaping half for all-unique labels (~5–6× less replay on that CAD pathology), but shipped **default-OFF** behind `perGlyphSubstitutedText`, gated on a browser real-font fidelity pass that has not landed. It delivers nothing to real users until flipped on.

### Acceptance

Page 0's `/DCTDecode` images decode off the main thread (build window no longer blocked on the browser codec), **and** the per-glyph substituted-text path is validated and enabled by default. Re-measure the build window on the #450 device trace afterward.

## Already landed

- **Diagnostics (the ticket's stated first step):** #455 split render timing into `wait` / `build` / `raster`; #457 split `build` into image-`decode=` and `replay=`; #514 split substituted-text `shape=` out of `replay`. Together these established that raster is a non-problem (~50 ms) and localized the real cost to decode + shaping.
- **Partial build-window fixes:** #588 (concurrent image decode, overlaps but doesn't off-thread), #517 (per-glyph shaping, default-off).
[...]

Issue #392, opened 2026-07-19
perf(cos): cache decoded stream bytes in CosDocument


ben-milanko:
## Summary

`CosDocument.decodeStreamData` re-runs the entire filter chain (and the cipher, on encrypted files) every time it is called. Nothing is cached on the `CosStream` or in the document. This is the single biggest performance lever in the repo: every consumer above pdf_cos re-decodes the same streams multiple times.

## Evidence

- `packages/pdf_cos/lib/src/document.dart:549-565` — `decodeStreamData` always re-decodes; only object *structure* is cached (`_cache`).
- `packages/pdf_cos/lib/src/filters/filters.dart:54-97` — `decodeStream` re-runs the full chain per call.
- Amplification:
  - Page render decodes every `/Contents` stream at least twice (image-collect pass + paint pass) — `packages/pdf_document/lib/src/page.dart:88-115` (`contentBytes()` also re-concatenates per call).
  - Form XObjects, annotation appearances, soft-mask forms re-decode per `Do` — `packages/pdf_graphics/lib/src/interpreter.dart:2292-2302`, `:775-823`, `:1651-1656`.
  - Text extraction, element enumeration, redaction, color processing each re-decode: `packages/pdf_document/lib/src/content_elements.dart:107`, `color_processing.dart:49,123`, `redaction.dart:148,166`, `vector_snapshot.dart:63-64`.
- Extra full-length copy per flate decode: `packages/pdf_cos/lib/src/filters/flate.dart:14` (`Uint8List.fromList` of an already-fresh buffer).

## Impact

Every page render, text search/selection, element hit-test, and content edit re-inflates multi-MB content streams 2-4+ times. On a 10 MB CAD content stream that is tens of ms of pure re-decode per pass; on encrypted files each pass also re-runs the cipher.

## Proposed fix

- Add a decoded-bytes cache in `CosDocument` keyed by `CosStream` identity (streams are identity-stable via the xref cache), with `stopBeforeFilter` as part of the key.
- Drop entries in `applyIncrementalUpdate` alongside `_cache`.
- Bonus: skip the extra `fromList` copy in `flate.dart` when the decoder already returns a `Uint8List`.
- Optionally cache the concatenated `contentBytes()` result per page-dict identity in pdf_document.

## Dependencies

None — this is the foundation ticket. #2 (parsed-ops cache), #3 (single-walk render), and the incremental-update and content-editing tickets all build on or benefit from it.

## Suggested order

[...]

PR #472, opened 2026-07-22
perf(editor): make the recorded single-walk render path the default (#394)

### LINKED ISSUE #394: perf(editor): make recorded single-walk render path the default
## Summary

Every page render interprets the content stream **twice**: a scan-only `ImageCollector` pass, then the paint pass. The scan pass still executes every operator — dispatch, operand reads, CTM math, per-glyph advance loops, form recursion. The single-walk `RecordingPdfDevice` path already exists and is proven pixel-identical, but is opt-in (`renderImageWithPlan` defaults `recorded: false`).

## Evidence

- `packages/dart_pdf_editor/lib/src/renderer.dart:108-135` — `renderPictureWithPlan`: parse once, `scanImagesOnly` walk (114-120), `decodeImages` (121), full paint walk (132-135).
- `packages/pdf_graphics/lib/src/interpreter.dart:953-977` — `_runOps`/`_runCursorOps` dispatch every op in scan mode; `_showText` runs fully (`:1742-1778`).
- Single-walk alternative: `renderer.dart:178-208` (`renderPictureRecordedWithPlan` via `RecordingPdfDevice`, whose `imageRequests` fuses image discovery into the paint walk); default off at `renderer.dart:405-414`.
- Workerless fallbacks still on the two-walk path: `pdf_page_view.dart:1637-1640` (deep-zoom detail), `editing_thumbnails.dart:2554` (thumbnails), `renderer.dart:612-640` (color sampler).

## Impact

On vector/text-dense pages with few images the scan walk is most of a full walk, so pages effectively interpret ~1.5-2x per render. Worst on the 100-420 ms/walk corpus pages. Nothing persists across re-renders either, so every zoom/page-color/annotation-toggle repaint re-walks.

## Proposed fix

1. Make the recorded single-walk path the only interpretation path (delete the collector+paint double walk).
2. Route the pdf_page_view deep-zoom fallback, thumbnail fallback, and color sampler through it.
3. Cache the parsed page op list (or recorded command buffer) per page+revision so re-renders replay instead of re-interpret.

## Dependencies

- Blocked by #392 (decoded-bytes cache) and #393 (ops cache) — land those first so the single remaining walk hits warm caches; deleting the scan pass first would double decode cost on uncached pages.

## Suggested order

Phase 1, ticket 3 of 16.

ben-milanko:
Closes #394.

The bitmap render path (`renderImage`/`renderImageWithPlan`) defaulted to `renderPictureWithPlan`, which interprets the page **twice**: a scan-only `ImageCollector` walk to discover images, then the paint walk. The record/replay path (`renderPictureRecordedWithPlan`) walks **once** — the `RecordingPdfDevice`'s `drawImage` calls double as image discovery — then replays a flat command buffer.

Per the #394 verification pass this is a **consolidation, not a 2×**: `scanImagesOnly` already short-circuits colour parsing, shading, path building, and glyph emission, so the scan walk is a fraction of the paint walk. But it removes a whole walk's operator dispatch + per-glyph advance cost and collapses the pipeline onto the single transcript the deep-zoom/tile route, `worker.record`, and the strip bins already consume — so it's well-exercised and parity-tested, not experimental.

## Changes

- `renderImage` / `renderImageWithPlan` default `recorded: false → true`.
- The two remaining direct `renderPictureWithPlan` production callers switch to the recorded variant: the page colour sampler (`PdfPageColorSampler.of`) and the workerless deep-zoom detail patch (`pdf_page_view.dart`).
- **Decoupled the strips device axis from `recorded`** — the gate was `deviceMode == strips && !recorded`; since `recorded` is now the default, that would have stopped the strip benchmark/parity harnesses (which set `deviceMode = strips` and call `renderImage` with the default) from engaging strips. Strips is an independent opt-in device mode, so the gate is now just `deviceMode == strips`. No production caller combines the two.

`renderPictureWithPlan` (two-walk) stays — it's what `render_record_replay_test` renders as the `direct` reference, and it's still reachable via `renderImage(recorded: false)`.

## Pixel-identity proof

- `render_record_replay_test` (zero-tolerance direct-vs-recorded): **pass**.
- `strip_parity_test`: **pass** (canvas reference is now the recorded path; strip-vs-canvas diffs unchanged).
- **Ghent render baselines** — the 54-PDF conformance suite renders through the now-recorded default and diffs against the two-walk baselines: **identical** to base (+40 pass, ~13 known colour deviations tolerated, GWG030 the one pre-existing spot/overprint failure — confirmed failing identically on the base commit). No new pixel diffs.
- `pdfjs_render_test` smoke (164): **pass**.
- Full `dart_pdf_editor` Flutter suite: green apart from GWG030 (pre-existing) and one worker-cancel **timing flake** (`render_worker_strips_test` "a late cancel id cannot abort the next worker request") that passes in isolation and in-cluster — my change doesn't touch worker routing.

## Measurement

The #394 comment asks for a `tool/perf.sh diff` on text-dense + vector-dense scenarios before quoting a number publicly. This is the correctness/consolidation half (proven pixel-identical) and makes no headline perf claim; the A/B is worth running before citing a figure.

Dev-log: `doc/dev-log/2026-07-22-recorded-render-default-394.md`.


ben-milanko:
## Pulling this back to draft — measured, and it's not a win

Per CLAUDE.md I should have A/B'd this before claiming a benefit. I did now, and the result argues against merging it as-is.

**A/B (flutter-render benchmark, Ghent corpus, scale 2, maxPages 2, repeat 3, headless engine, this branch vs `main`):**

```
main  (two-walk):  73.62 ms/page
#394 (recorded):   74.70 ms/page   (+1.5%)
```

The per-file split is mechanistically consistent, not noise:
- **Image/ICC-heavy pages regress** (ICC-CMS `+49.6 ms`, GWG060 shading `+8.6 ms`, GWG161 `+8.3 ms`) — the record→replay round-trip re-issues every image/paint command, and the `scanImagesOnly` walk it removes was already cheap.
- **Vector/spot pages improve, but only slightly** (SPOT `−3.2 ms`, GWG132 `−2.6 ms`).

More importantly, real-session devtools traces (Windows, 28-page editing session) show the dominant render cost is the **UI-thread picture build from the worker record — up to 1010 ms/frame (#454)** — with raster ~2 ms. Making `recorded` the default routes *more* rendering through that same record→replay build, i.e. the wrong direction until #454's build cost is understood and fixed.

Keeping the branch as a draft/reference. The correctness work stands (proven pixel-identical across Ghent + pdfjs + zero-tolerance parity), but it should not merge as a perf change. Reassess after #454.

ben-milanko:
## Rebased onto current `main` (includes #392) + validation

Merged latest `main` in — this is now unblocked: #392 (decoded-stream cache, #481) is merged, so the single remaining interpretation walk hits the warm decode cache, satisfying the #394 dependency ("land the caches first").

### Correctness (pixel-identical)
- `render_record_replay_test` (record==direct parity), `knockout_group_test`, `mesh_shading_test`, `benchmark_paint_test` — all green.
- `dart analyze` clean.

### Perf evidence (real-Chrome `tool/perf.sh`)
Two regimes, because this flips the **local** render default and the web default has the worker on:

- **Worker on** (default web) — `webdiff main scroll-plan`: the render worker already interprets via its own single-walk transcript (`interpret paths worker=19 recorded=0`), so this change is a **no-op on the hot path** — every metric within noise, **✓ no regression** (buildP50 0.92→0.93, buildP95 1.36→1.36, jank 7→7).
- **Worker off** (`PERF_NO_WORKER=true`, the default for viewer-only web apps post-#465, plus thumbnails/color-sampler/deep-zoom fallbacks): the local path now takes the recorded single-walk (`interpret paths recorded=2`, previously the two-walk collector+paint). This is where the removed scan walk saves work. A clean A/B number isn't available because the harness's `NO_WORKER` control logs worker-404 fallback lines it counts as run failures — a harness limitation, not a regression.

### Framing
This is best understood as a **consolidation** (per the original body): one interpretation path instead of two, pixel-identical, regression-free on the worker path, with the measurable saving on the workerless/thumbnail/fallback paths. Not oversold as a hot-path win.

**Follow-up:** the display-list `renderPicture` callers (thumbnail display-list, overlay previews) still use the two-walk path and could be routed similarly in a later change.

PR #503, opened 2026-07-22
fix(render): restore blend mode on form/appearance exits (#462)

### LINKED ISSUE #462: CanvasPdfDevice leaks blend mode across forms/annotations (viewer renders opaque content multiplied)
## Summary

`CanvasPdfDevice._blend` (dart_pdf_editor `canvas_device.dart`) is a plain field that is **not** saved/restored across `save()`/`restore()`. Because the interpreter brackets every form, annotation appearance, and `q`/`Q` with `device.save()/restore()` but restores its *own* graphics state on the way out **without** re-issuing `setBlendMode`, a non-Normal blend set inside a form (e.g. a Multiply highlight) leaks out and applies to whatever is painted next.

Result: a Normal-blend object drawn after a Multiply one renders **multiplied against the backdrop** in the viewer instead of opaque.

## Why it's mostly unnoticed

Real files where this bites need a non-Normal blend inside a form/appearance followed by more painting that expects Normal — e.g. a Multiply highlight annotation followed by an opaque annotation (stamp/ink/square). A page of all-highlights hides it because everything is Multiply anyway.

## Root cause

`CanvasPdfDevice.save()` → `canvas.save()`, `restore()` → `canvas.restore()`. Those save the *canvas* graphics state, but `_blend` is a Dart field, so it is not part of that state and survives the restore.

The interpreter's `Q` handler re-syncs the device (`interpreter.dart:1051-1052`, `device.setBlendMode(restored.blendMode)`), but the **form/appearance exits do not**:
- `_drawAppearance` restores `_state = savedState` then `device.restore()` (~`interpreter.dart:836-837`)
- `_drawForm` restores `_state = _stateStack.removeLast()` then `device.restore()` (~`interpreter.dart:2341-2342`)

Neither re-issues `setBlendMode`, so the device's `_blend` stays whatever the form last set.

## Reproduction (conceptual)

Page content: `Do` a form that does `/GS0 gs` (BM Multiply) + fill; after the form returns, page content fills again with Normal blend. The second fill renders multiplied. (Confirmed on the analogous vector-print device while working on #461 — a Normal-blend fill after a Multiply form came out translucent.)

## Suggested fix

Root-cause it in the interpreter: at the form/appearance (and tiling-pattern / soft-mask-form) state-restore sites, re-sync the device blend the way `Q` already does — e.g. `if (_state.blendMode != restored.blendMode) device.setBlendMode(restored.blendMode);` before `_state = restored`. That fixes every device at once.

Alternatively, mirror `_blend` on a save/restore stack inside `CanvasPdfDevice` (the approach taken for the vector-print device in #461 — `save()` pushes, `restore()` pops). The interpreter fix is preferable since it also covers any other device with the same shape.

## Scope / relation to #461

Pre-existing and latent; **not** introduced by #461. PR #461 fixed the equivalent leak in the Windows/Linux vector-print device (it surfaced there because that device newly tracks blend). This issue tracks the viewer-side (`CanvasPdfDevice`) counterpart, which wants a Flutter render-level regression test.

ben-milanko:
## Summary

Fixes #462. A non-Normal blend mode set inside a form XObject / annotation appearance / transparency group / tiling pattern **leaked** out and applied to whatever was painted next — an opaque object drawn after a Multiply one rendered multiplied against the backdrop.

The interpreter brackets each of these scopes with `device.save()`/`restore()` and restores its own graphics state on the way out, but a device's blend mode is **not** part of the canvas save stack (`CanvasPdfDevice` tracks it in a plain Dart field), so `device.restore()` doesn't undo it. The `q`/`Q` handler already re-issues `setBlendMode` when the restored state's blend differs; the five form/appearance/group/pattern exit sites did not.

## Change

- Add `_restoreDeviceBlend()` (mirrors the `Q` handler) and call it at all five exits: `_drawAppearance`, the soft-mask form, the transparency-group form, the tiling-pattern, and `_drawForm`.
- Fixing it in the **interpreter** (not just `CanvasPdfDevice`) covers every device at once — `CanvasPdfDevice`, the vector-print device, `RecordingPdfDevice` — and matches where `Q` already does it, so the **recorded and painted paths stay identical**.

~22 LOC in `interpreter.dart`.

## Testing (evidence)

New `blend_leak_test.dart`: a form sets `/BM Multiply` at its **top level** (no inner `q`/`Q`, so only the form *exit* can restore it) and fills, then an opaque white rect is drawn over a red backdrop. White × red = red under Multiply, so the "white" region reads **green channel 0 without the fix, ~255 with it**.

- Verified **failing without** the change (`Expected: > 200, Actual: 0`), **passing with** it.
- `pdf_graphics` suite (884) green.
- `render_record_replay_test` (record==paint equivalence), `knockout_group_test` (transparency groups), `mesh_shading_test` green.

## Notes

Correctness fix, not perf — not `perf.sh`-measurable; the golden-style pixel test is the evidence, as the issue requested. The analogous vector-**print** blend leak was fixed under #461; this is the viewer/render-side counterpart.


ben-milanko:
Rebased onto current main and resolved the conflicts (blend fix is preserved and now verified against main's structure).

Two conflicts, both in `interpreter.dart`:
- **`_drawForm` exit** — main had independently added the same blend re-sync *inline*; took this PR's `_restoreDeviceBlend(restored)` helper (equivalent, and consistent with the helper used at the other four sites).
- **Type3 glyph exit** — main's #535 record-once refactor moved the Type3 state isolation out of the `execute` callback and into `_executeType3Proc`, so this PR's block there was stale. Dropped it and re-applied `_restoreDeviceBlend(savedState)` inside `_executeType3Proc` where the state restore now lives — so a CharProc's blend still can't leak, and because the reset is captured while recording a cell, a *stamped* glyph can't leak it either.

Verified on the resolved tree: `blend_leak_test` green (the golden evidence), `render_record_replay_test` green (record==paint parity intact), full `pdf_graphics` suite (923) green, bundle regenerated, root analyze clean.

PR #513, opened 2026-07-23
perf(web): browser JPEG decode — observability + main-thread fast path (#458)

### LINKED ISSUE #458: perf(web): the worker's browser JPEG decode silently declines, so the main thread pays 643 ms
## Summary

The web render worker **already** decodes JPEGs with the browser's own codec (`_withBrowserDecodedImages` → `_decodeWithBrowserCodec`, `render_worker_web_entry.dart`), and it is wired into the record path. On the reported document it **declines to**, and the main thread pays 643 ms instead.

## Evidence

From a device trace, the image-bearing record for page 0:

```
[perf 2760] phase worker=0 kind=record page=0 ... decode=1.0ms serialize=17.0ms ... transcript=hit
[perf 3429] interpret page=0 FIRST interpret=3361.2ms wait=526.0ms build=668.7ms decode=643.4ms replay=24.7ms
```

- The worker's browser-codec pass ran and took **1.0 ms** for a page with two `/DCTDecode` images — i.e. it declined both, immediately.
- The record is **888,812 B**. Had the worker decoded the 1280×1656 cover, it would have carried ~8.5 MB of RGBA.
- The main thread then spent **643.4 ms** decoding them through `ui.instantiateImageCodec` (Skia's WASM codec, on the UI thread under CanvasKit).

So the capability exists, is invoked, and silently no-ops.

## Most likely cause

`_decodeWithBrowserCodec` returns null early when:

```dart
bool get _browserImageDecodeAvailable =>
    globalContext.has('Blob') &&
    globalContext.has('createImageBitmap') &&
    globalContext.has('OffscreenCanvas');
```

A false here declines every image at near-zero cost, which matches `decode=1.0ms` for two images exactly. The other early exits are content-dependent (`/ImageMask`, `DeviceCMYK` without a DCT soft mask) and would not both fire on this page — its images are non-CMYK DCT, which is precisely the case the browser path exists to serve.

## First step: make the guard observable

The worker's `ready` line reports `sharedBytes`; it should report this too, e.g.

```
webworker ready worker=0 ... sharedBytes=true browserImageDecode=true
```

and `_decodeWithBrowserCodec` should log its decline reason once per page under the perf log. Without that, a silently-disabled decode path is indistinguishable from one that ran, which is what cost this investigation a full detour.

If the flag turns out true, the next suspect is the per-image early exits and they need the same treatment.

## Why this matters more than it looks

This is the difference between 643 ms on the UI thread and ~0. It is also a reminder that the existing capability was invisible: I began implementing a **second** browser-codec decode path before discovering the first one, because nothing in any trace indicated it existed or that it had opted out.

ben-milanko:
Closes #458.

The web render worker already decodes `/DCTDecode` images with the browser's native codec, but on the reported mobile-web trace (24 MB / 62-page book) it **silently declined** and the main thread paid **643 ms** in Skia's WASM codec (`ui.instantiateImageCodec` under CanvasKit). Nothing distinguished a codec that ran from one that no-opped. Two independent pieces here.

## Part 1 — make the decline observable
`decode=1.0ms` for two DCT images can only mean the worker's capability guard returned false (a real `createImageBitmap` decode awaits and costs real time).

- The worker `ready` line now reports `browserImageDecode=<bool>` and, when false, `missing=<Blob|createImageBitmap|OffscreenCanvas>` — one line pins the cause per worker.
- Per page, a decline tally folds a summary (`codec=2` / `codec=0 declined=2(noCapability)`) into the phase log next to `decode=`, distinguishing the #458 `noCapability` case from the expected `notDct`/`cmyk`/`imageMask` pure-Dart declines. Rides on `PdfRenderTrace.imageDecodeSummary` (web-only, carried like `cosStats`); zero overhead when timings are off.

## Part 2 — main-thread browser-codec fallback (the latency win)
The near-certain missing prerequisite is **`OffscreenCanvas` in the worker scope** (absent on iOS Safari <16.4), which the worker needs to read pixels back. The **main thread doesn't**: `dart:ui_web`'s `createImageFromImageBitmap` consumes an `ImageBitmap` straight into a GPU `ui.Image`, no canvas, no readback.

- New conditional-import seam `browser_jpeg_decode.dart` (stub + `_web`). `decodeJpegWithBrowser` uses `createImageBitmap(blob, {resize…})` (resize preserves the display-res decode cap) → `createImageFromImageBitmap`; returns null off-web/on failure.
- `image_decoder.dart`'s `_decodeOne` tries it first, falling back to the engine codec only when null. Mask/`/Decode`/color-key handling unchanged (browser base is opaque → readback math identical).

Native Flutter is untouched (stub returns null; the engine codec already runs off the raster thread there).

## Verification
- `dart analyze` clean; worker still compiles under `dart compile js`.
- `flutter build web` compiles the main-thread `_web` file (+ Wasm dry run passed) — catches the dart2js-only breakage the analyzer misses.
- `render_trace_gate` + `render_worker` + `image_decoder` + `editing_image` + `inline_image` tests pass.

## Not in this PR
- Regenerating the committed `pdf_render_worker.dart.js` asset (the flaky auto-regen path, #422).
- Confirming *which* prerequisite is missing on the reported device — Part 1 now makes that a one-line trace read.


ben-milanko:
### Web A/B (`tool/perf.sh webdiff main scroll-scan`)

Ran the image-heavy scroll scenario (12p A3 scan) — the closest workload that exercises the new per-image worker tally (its 24 Flate images/page each decline as `notDct`).

**7 iterations — clean pass, no metric regressed past 1.10×:**

| metric | baseline | current | Δ |
|---|---|---|---|
| buildP50 | 1.26 | 1.22 | −4.0% |
| buildP95 | 3.23 | 3.33 | +3.1% |
| buildMax | 15.75 | 16.07 | +2.0% |
| buildOver50 | 0 | 0 | — |
| jankCount | 11 | 8 | −27% |
| openPageCountMs | 182 | 169 | −7.1% |
| agentMemoryBytes | 290.0M | 290.2M | +0.0% |

(A first 3-iteration run flagged `jankCount` 10→12; the 7-iteration re-run reversed it to 11→8 — variance on a small-integer frame count, and mechanistically impossible for these changes to cause: Part 1's tally runs in the worker isolate, off the UI thread.)

**Scope note:** the desktop-Chrome harness cannot demonstrate Part 2's actual win. Its worker scope *has* `OffscreenCanvas`, so `/DCTDecode` images decode in the worker and the new main-thread fallback never fires; and the corpus has no non-CMYK DCT scenario to point it at anyway (`worker decode max=0.00MB` confirms no image bytes crossed the boundary here). The 643 ms → browser-codec win only manifests where the worker *declines* (e.g. iOS Safari <16.4), which needs a device trace to measure — Part 1's `ready`-line reporting now makes that a one-line read.

Verified: no regression on the common image path.

ben-milanko:
### Desktop A/B for the main-thread decode (Part 2's win, now measurable)

Follow-up to the earlier `scroll-scan` run, which couldn't exercise Part 2 (desktop worker has `OffscreenCanvas`; no non-CMYK DCT scenario existed). Added the tooling to force the path:

- **`photo-jpeg-6p.pdf`** — 6 full-page DeviceRGB DCTDecode covers (~1240×1654, the #458 shape), byte-deterministic via `package:image` `encodeJpg`. The other 8 corpus files stay byte-identical.
- **`?worker=0`** harness toggle → nulls `pdfRenderWorkerScriptUrl`, forcing interpret + decode onto the main thread so every JPEG runs through the changed `_decodeOne`. The toggle lives in the harness (copied into the baseline worktree), so both trees fork at the same lever.
- **`scroll-photo-noworker`** scenario.

**`tool/perf.sh webdiff main scroll-photo-noworker --iterations 5` — clean pass:**

| metric | baseline (Skia) | current (browser codec) | Δ |
|---|---|---|---|
| **buildP50** | **1.49** | **1.22** | **−17.8% ✓** |
| buildMax | 14.71 | 14.51 | −1.4% |
| openPageCountMs | 232 | 219 | −5.3% |
| jankCount | 1 | 1 | — |
| agentMemoryBytes | 138.5M | 138.4M | −0.1% |

In `worker=0` mode the *only* code difference between the trees is `_decodeOne`'s DCT branch, so the consistent −17.8% P50 **is** the browser codec working and beating Skia's WASM decode on the UI thread — had it silently fallen back, current would equal baseline. Part 2 confirmed active + faster, not just compiling.

#### Bonus finding from Part 1's tally
Running the same doc with the worker **on** (headless), the worker's browser decode declines every page: `imageDecode=codec=0 declined=1(decodeNull:1)` — `createImageBitmap`/`OffscreenCanvas.getImageData` returns null under headless Chrome (the capability getter is *true*, so it's not `noCapability`). Part 1 surfaced this instantly. Whether it also bites real mobile browsers is a #458 sub-question for a device trace. (Renamed that reason `decodeFailed`→`decodeNull` in this push: the phase log rides the harness's fatal-error scan and "fail" was tripping it.)

> Note: the `decodeNull` rename touches the worker lib, so the bundled `pdf_render_worker.dart.js` (auto-regenerated at 74cdefc8) is stale again — leaving it to the #422 auto-regen rather than hand-rolling it.

PR #514, opened 2026-07-23
perf(editor): split substituted-text shaping out of the replay phase (#454)

ben-milanko:
Refs #454.

#454's replay-bound half (dense vector/CAD pages: ~877 ms of pure `ui.Picture` construction, ~0 decode) was attributed to *"raw `ui.Canvas` call volume plus `TextPainter` shaping of mostly-unique CAD labels"* — but which dominates was never measured. This adds the split, the same disciplined step that #455/#457 used to attribute the build phase, **before** anyone picks a fix (this ticket's "certain" cause has been overturned four times by measurement).

## The split
`CanvasPdfDevice` shapes substituted-font runs via `TextPainter` in `_measureLayout`, cached by full run text (2048 LRU). The cache-miss factory is the shaping; a hit is a cheap lookup. So:
- `_measureLayout`, when `PdfPerfLog.enabled`, times only the miss factory and counts miss/hit (`_shapeLayout` extracted so the hot path is unchanged and pays nothing).
- `pdf_page_view` resets the accumulators around the retained-scene replay and threads them in.
- `PdfPerfLog.interpret` prints ` shape=<ms> shaped=<miss> cached=<hit>` beside the existing `decode=`/`replay=` split.

## What it measured (desktop web harness)
| scenario | page | replay | shape | shaped/cached |
|---|---|---|---|---|
| scroll-plan | 2 | 23.6 ms | **0.0 ms** | 0 / 417 |
| scroll-plan | 5 | 57.7 ms | **20.3 ms** | 362 / 0 |
| scroll-plan | 7 | 54.2 ms | **22.3 ms** | 397 / 0 |
| scroll-diagram | 2 | 84.3 ms | **0.0 ms** | 0 / 1338 |

**Finding — shaping's share is gated on label uniqueness, and that reframes the fix:**
- **Repeated labels** (plan 2–4, the whole diagram) → the run-cache absorbs them, `shape≈0`, replay is 100% raw canvas-call volume.
- **Unique labels** (plan 5–7 — exactly #454's "mostly-unique CAD label" condition) → 100% miss, shaping ≈ **37% of replay**; the other ~63% is still canvas calls.

So a per-glyph cache (the natural fix for unique labels) is worth up to ~37% of replay *on all-unique pages* and ~0 elsewhere — **not** a fix for the vector-dense diagram case, which is canvas-call-volume bound with shaping already at zero. Command-count / slug-layer work is the bigger lever there.

Measurement-only; no behaviour change. VM text/paint/perf-log tests green; analyzer clean.


ben-milanko:
### Worst-case measurement: `scroll-cad-labels` (both levers quantified)

Added `cad-labels-6p.pdf` — A3 sheets of **660 survey-point labels, unique across the whole document** (substituted Helvetica), so the run-cache misses on every one. This reproduces #454's "mostly-unique CAD label" pathology in pure form (the existing scenarios' labels repeat, so they can't).

```
page 2  replay=55.7ms  shape=37.1ms  shaped=660  cached=0    <- first render, all unique
page 4  replay=52.3ms  shape=37.0ms  shaped=660  cached=0
page 5  replay=48.0ms  shape=33.0ms  shaped=660  cached=0
page 5  replay= 3.8ms  shape= 0.0ms  shaped=0    cached=660  <- re-render, all cached
```

**Both levers, on the worst case:**

| lever | size | what it is |
|---|---|---|
| **Text cache** | **~92% of replay** (48 ms cold → **3.8 ms warm**, ~13×) | first-encounter text work: 33 ms measured layout (`shape=`) + ~11 ms first-paint/raster the clock doesn't cover |
| **Canvas-call floor** | **~3.8 ms** (~8% here) | the 660 draw calls; what the warm + vector-dense cases are bound by |

**Conclusion for the fix:** a **per-glyph** cache (turning unique *runs* into cached *glyphs*) is strongly justified **for the unique-label pathology** — it could push a cold unique-label page from ~48 ms toward the ~4 ms floor (~10×) — and is **irrelevant elsewhere** (warm pages and the vector-dense diagram are canvas-call-volume bound, shaping already 0). It's a large cross-layer change (interpreter → codec → device, Ghent/pdf.js baseline gated), but it now has a **measured ceiling and a reproducible cold-vs-warm A/B target** (`scroll-cad-labels`).

ben-milanko:
## Self-review + fixes (rebased)

Reviewed the diff; three findings, all addressed in this push:

1. **Unused `seed` parameter** on `buildCadLabels` — the labels are counter-derived (no RNG), so `{seed}` was dead and misleading. Removed; `cad-labels-6p.pdf` regenerates **byte-identical**.
2. **No test for the new log field** — added a `perf_log_test` case asserting the `interpret` line's `shape=/shaped=/cached=` split appears when measured and stays off when not, matching the existing `wait/build` and `decode/replay` split tests.
3. **Was `CONFLICTING`** — #513 merged after this branched, so both PRs touched `gen_public_corpus.dart` (the rebase interleaved the two structurally-similar builders) and `scenarios.json`. Rebased onto current `main`, reconstructed both `buildPhotoJpeg` (from #513) and `buildCadLabels` as separate clean functions, and kept both `write()` calls and both scenarios. Verified: both corpus PDFs regenerate byte-identical, analyzer clean, `perf_log_test` + `text_cache_test` green.

**Correctness re-confirmed:** the static shaping accumulators are reset→read across a *synchronous* `scene.replay()` (no `await` between), so nothing interleaves; instrumentation is fully gated on `PdfPerfLog.enabled`; `_shapeLayout` extraction is behaviour-identical to the old inline factory.

Now `MERGEABLE`.

PR #517, opened 2026-07-23
perf(editor): per-glyph substituted-text composition (#454)

ben-milanko:
Refs #454. The optimization for #454's replay-bound half, measured and validated.

## Problem
`CanvasPdfDevice` shapes non-embedded-font runs via `TextPainter`, cached by full run text. Unique labels (CAD survey points, coordinate grids) miss that run cache on **every** render and re-shape — the measured worst case was a cold page's `replay` at ~48 ms vs ~4–8 ms warm (the split PR #514 quantified this).

## Approach
Compose such a run from **per-character** layouts: each character is shaped once into a new `_glyphCache`, and a run is placed from those cached glyphs at cumulative advances, then scaled to the PDF's total width exactly as the whole-run path is. A page of 660 unique labels collapses to an alphabet of ~14 cached glyphs.

## Result (`scroll-cad-labels`, `?perGlyph=1` vs default, same build)
| page | perGlyph=0 (whole-run) | perGlyph=1 (composed) |
|---|---|---|
| cold 2 | replay=39.5ms shape=26.9 | **replay=8.0ms** |
| cold 4 | replay=48.1ms shape=34.1 | **replay=8.1ms** |
| cold 5 | replay=50.8ms shape=34.4 | **replay=15.1ms** |

**~5–6× less replay** on the exact pathology (48 → 8 ms), landing near the warm floor.

## Fidelity — why it's gated and default-off
Per-character placement drops cross-character kerning and any contextual shaping, so `_composableRun` restricts it to a **safe set**: ASCII digits, uppercase letters, space, and technical punctuation — none join or ligate, and the digits (the bulk of CAD labels) are tabular. Lowercase (fi/fl/ff ligatures) and non-ASCII (Arabic/CJK/combining) fall back to whole-run shaping, as do stroked/gradient runs.

Behind `CanvasPdfDevice.perGlyphSubstitutedText`, **default off**. The mechanics are pixel-exact under the (no-kern) test font — `per_glyph_text_test` asserts composed == whole-run byte-identical — but **real-font uppercase kerning can only be judged in a browser against the Ghent/pdf.js renders**. That browser fidelity pass is the precondition for flipping the default; until then it ships as a proven, validated opt-in.

## Validation
- `per_glyph_text_test`: (1) composable run composes pixel-identically to whole-run, (2) non-composable (lowercase) run untouched by the flag, (3) 200 unique runs → run cache empty, glyph cache < 20 (an alphabet).
- **No regression default-off**: byte-identical substituted-text render test passes; the 20 Ghent failures are pre-existing (CMYK/overprint/softmask — verified identical on a clean tree), not from this change.
- `flutter build web` compiles the device + harness change.

## Incidental fix
The run-cache key string had its two literal spaces stored as **NUL bytes** (`\x00`) — pre-existing corruption (harmless: NUL is still a valid separator, but wrong). Fixed to real spaces.

## Not in this PR
The browser/pdf.js real-font fidelity pass and the default-on flip — the natural follow-up, gated on that evidence.

### DIFF
--- a/app/tool/perf/driver.mjs
+++ b/app/tool/perf/driver.mjs
@@ -311,6 +311,7 @@ async function main() {
   if (process.env.PERF_QUERY) qp.set('query', process.env.PERF_QUERY);
   if (process.env.PERF_REPEAT) qp.set('repeat', process.env.PERF_REPEAT);
   if (process.env.PERF_OPS) qp.set('ops', process.env.PERF_OPS);
+  if (process.env.PERF_PER_GLYPH) qp.set('perGlyph', process.env.PERF_PER_GLYPH);
   const qs = qp.toString();
   const url = `http://127.0.0.1:${PORT}/${qs ? '?' + qs : ''}`;
   if (SCENARIO) console.log(`▶ scenario ${SCENARIO.name} (${SCENARIO.kind}) pdf=${SCENARIO.pdf}`);
--- a/app/tool/perf/perf_harness.dart
+++ b/app/tool/perf/perf_harness.dart
@@ -121,6 +121,11 @@ final int _imageCacheMb = _qInt('imageCacheMb', 0);
 /// path never runs. Default 4 (the app's normal pool).
 final int _workerPool = _qInt('worker', 4);
 
+/// `?perGlyph=1` turns on per-glyph substituted-text composition (#454) so the
+/// same build can A/B it against the whole-run shaping default - watch the
+/// interpret line's `replay=`/`shape=` on `scroll-cad-labels`.
+final bool _perGlyph = _qBool('perGlyph', false);
+
 // ---------------------------------------------------------------------------
 // Capture: every debugPrint line + every frame's timing + scenario metrics.
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


ben-milanko:
## Fidelity pass done → gate tightened → **default on**

Ran the real-Chrome fidelity probe (the VM/test-font environment has no kerning, so it can't measure this): whole-run `fillText` vs per-character `measureText` advances, both scaled to one box on Helvetica/Arial/Courier — exactly what the device does — pixel-diffed.

**First finding:** the original gate (all uppercase) was too loose. Realistic all-caps words diverge substantially because kerning fires between adjacent letters:

| string | changed pixels |
|---|---|
| PAY ATTENTION | 57% |
| TAX INVOICE | 38% |
| AVENUE | 31% |
| WATER TANK | 27% |
| WARNING | 17% |
| N1234567.89 E7654321.01 | **0%** |
| REVISION A3 SHEET 12 OF 48 | **0%** |

**Fix:** tightened `_composableRun` so a letter may only neighbour a digit or a space (never another letter or punctuation) — exactly the shape of coordinate/survey/part labels. Re-probed:

```
MAX diff among gate-passing strings: 0%   (Helvetica, Arial, Courier)
```

Every string the gate composes is **pixel-identical** to whole-run; every kerning-pair word falls back to whole-run shaping. It's deliberately conservative (safe adjacent-letter strings like `SHEET`/`DN200` also fall back — the gate can't cheaply know they don't kern; a future per-font kerning table could widen it).

**With the pixel diff proven nil on real fonts, `perGlyphSubstitutedText` now defaults ON.**

- **Win survives the tighter gate:** `scroll-cad-labels` cold-page replay **~52 → 7.5 ms (~7×)** — coordinate labels still compose.
- **No baseline shift:** VM text tests pass; the Ghent suite fails the same **20 pre-existing** tests as a clean tree (test font has no kerning → composed == whole-run there anyway).

This supersedes the "default-off, needs a browser fidelity pass" note in the PR description — the pass is done and it's on.

PR #538, opened 2026-07-23
feat(editor): custom menu support

SpanManX:
## Why

The built-in right-click / long-press menu is fixed UI: a fixed set of items,
one language. Apps need custom menus — localized copy, app-specific actions
("Send to…", "Open in editor"), different items per page or annotation type.
This lets the host fully own the menu when the default one doesn't fit.

## What

Adds `PdfViewer.onAnnotationMenuRequested`. The default menu keeps owning
the gesture as before; this callback is the host's escape hatch — set
`contextMenuEnabled: false` and supply the callback, and the viewer hands
the gesture to you instead of dropping it.

## Affected packages

- [ ] `pdf_cos`
- [ ] `pdf_document`
- [ ] `pdf_graphics`
- [x] `dart_pdf_editor`
- [ ] `pdf_test_fixtures`
- [x] Example app

## Change type

- [ ] Bug fix
- [x] Feature
- [ ] Rendering change
- [ ] Editing behavior change
- [ ] Performance change
- [ ] Refactor / cleanup
- [x] Tests / fixtures only
- [ ] Documentation only

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze`
- [x] `cd packages/pdf_cos && fvm dart test`
- [x] `cd packages/pdf_document && fvm dart test`
- [x] `cd packages/pdf_graphics && fvm dart test`
- [x] Real PDF corpus parse/render smoke test
- [x] Example app smoke test

## Compatibility checklist


ben-milanko (commented):
Thanks for this — the design is a clean escape hatch and it's nicely backward-compatible (null callback makes every new path a no-op, so existing `contextMenuEnabled: false` users are unaffected). I checked it out locally: `dart analyze` is clean and the existing menu suites (`editing_longpress_menu_test.dart`, `pdf_viewer_test.dart`) still pass. A few things I'd like to see addressed before merge:

### 🔴 Reader-mode touch long-press on an annotation is silently dropped
The callback is wired into 3 of the 4 menu entry points — desktop right-click (`_onSecondaryTapUp`), the editing-overlay long-press (`_onMenuLongPress`), and touch text-selection (`_onLongPressStart`) — but not `_maybeAnnotationMenu`, the reader-mode touch path for an annotation with no selectable text under it:

```dart
bool _maybeAnnotationMenu(LongPressStartDetails details) {
  final editing = widget.editing;
  if (editing == null || editing.tool != null || editing.isPickingColor ||
      !widget.contextMenuEnabled) {   // host-takeover mode returns false here
    return false;
  }
```

So on mobile/reader mode, long-pressing an annotation with the stock menu disabled returns `false` → `_clearSelection()`, and the host callback never fires. That's exactly the "app-specific actions / different items per annotation type" scenario from the PR description, on a primary mobile surface — this path needs the same host-takeover branch the other three got.

### 🔴 No tests for the new public API
All six changed files are lib/example. The repo already has the harness for this — `longPressAt` / `viewPoint` under `group('contextMenuEnabled: false')` in `editing_longpress_menu_test.dart`. Could you add coverage that: the callback fires with the right `(pageIndex, pagePoint)` for each entry point; it is *not* called when `contextMenuEnabled: true`; and a null callback drops the gesture cleanly? The Validation checkboxes (`flutter test`, Ghent, corpus, example smoke) pass, but nothing added exercises the new behavior, so they're really "build still green" rather than "feature verified."

### 🟡 Name says "Annotation," behavior is "any context menu"
`onAnnotationMenuRequested` / `PdfAnnotationMenuHost` also fire for plain-text right-click and text long-press selection. Since this is new public API, the annotation-specific name is a future breaking-change trap — `onContextMenuRequested` would match what it actually does. Better to settle it now than rename later.

### 🟡 Callback carries no annotation identity
It hands over `(globalPosition, pageIndex, pagePoint)` only, so a host wanting per-annotation menus has to re-hit-test. That's doable via `editing.selectableAnnotationAt(page, x, y)` on the controller the host already holds — but it's undocumented. Either mention that pattern in the doc-comment or pass the hit result through.

### 🟢 Minor
- Nice, idiomatic demo — the `showMenu` root-overlay anchoring is the standard recipe.
- Worth a quick confirm (not a proven bug): when selectable text *and* an annotation both sit under one long-press, that only one callback invocation results (the text path and the overlay annotation path shouldn't both fire).

Happy to look again once the reader-mode path and tests are in.

SpanManX:
> Thanks for this — the design is a clean escape hatch and it's nicely backward-compatible (null callback makes every new path a no-op, so existing `contextMenuEnabled: false` users are unaffected). I checked it out locally: `dart analyze` is clean and the existing menu suites (`editing_longpress_menu_test.dart`, `pdf_viewer_test.dart`) still pass. A few things I'd like to see addressed before merge:
> 
> ### 🔴 Reader-mode touch long-press on an annotation is silently dropped
> The callback is wired into 3 of the 4 menu entry points — desktop right-click (`_onSecondaryTapUp`), the editing-overlay long-press (`_onMenuLongPress`), and touch text-selection (`_onLongPressStart`) — but not `_maybeAnnotationMenu`, the reader-mode touch path for an annotation with no selectable text under it:
> 
> ```dart
> bool _maybeAnnotationMenu(LongPressStartDetails details) {
>   final editing = widget.editing;
>   if (editing == null || editing.tool != null || editing.isPickingColor ||
>       !widget.contextMenuEnabled) {   // host-takeover mode returns false here
>     return false;
>   }
> ```
> 
> So on mobile/reader mode, long-pressing an annotation with the stock menu disabled returns `false` → `_clearSelection()`, and the host callback never fires. That's exactly the "app-specific actions / different items per annotation type" scenario from the PR description, on a primary mobile surface — this path needs the same host-takeover branch the other three got.
> 
> ### 🔴 No tests for the new public API
> All six changed files are lib/example. The repo already has the harness for this — `longPressAt` / `viewPoint` under `group('contextMenuEnabled: false')` in `editing_longpress_menu_test.dart`. Could you add coverage that: the callback fires with the right `(pageIndex, pagePoint)` for each entry point; it is _not_ called when `contextMenuEnabled: true`; and a null callback drops the gesture cleanly? The Validation checkboxes (`flutter test`, Ghent, corpus, example smoke) pass, but nothing added exercises the new behavior, so they're really "build still green" rather than "feature verified."
> 
> ### 🟡 Name says "Annotation," behavior is "any context menu"
> `onAnnotationMenuRequested` / `PdfAnnotationMenuHost` also fire for plain-text right-click and text long-press selection. Since this is new public API, the annotation-specific name is a future breaking-change trap — `onContextMenuRequested` would match what it actually does. Better to settle it now than rename later.
> 
> ### 🟡 Callback carries no annotation identity
> It hands over `(globalPosition, pageIndex, pagePoint)` only, so a host wanting per-annotation menus has to re-hit-test. That's doable via `editing.selectableAnnotationAt(page, x, y)` on the controller the host already holds — but it's undocumented. Either mention that pattern in the doc-comment or pass the hit result through.
> 
> ### 🟢 Minor
> * Nice, idiomatic demo — the `showMenu` root-overlay anchoring is the standard recipe.
> * Worth a quick confirm (not a proven bug): when selectable text _and_ an annotation both sit under one long-press, that only one callback invocation results (the text path and the overlay annotation path shouldn't both fire).
> 
> Happy to look again once the reader-mode path and tests are in.

OK — actually I'm planning a rewrite. Hit some gaps in real-project usage that the current shape doesn't fully cover for the customization I need.

ben-milanko:
Thanks for the rewrite — I re-reviewed the current head (`628a69f`). The earlier naming, reader-mode annotation path, and test-coverage concerns are addressed, but I found three remaining issues that I think should be fixed before merge:

### 1. Desktop takeover runs before the clicked context is prepared

In `_onSecondaryTapUp`, the `contextMenuEnabled: false` branch invokes `onContextMenuRequested` and returns before the stock path:

- selects the right-clicked word,
- selects the hit annotation,
- or selects a form widget.

That makes desktop behavior inconsistent with the touch takeover paths, which do prepare selection context first. I reproduced this with a diagnostic widget test: right-clicking the word “Page” produced `controller.selectedText == ''` inside the callback, whereas the stock menu selects `"Page"`. Consequently the example's **Copy (custom)** action has no text to act on unless the user happened to select text beforehand.

Please either perform the same target/selection preparation before the handoff, or pass the resolved context through the callback request.

### 2. Empty long-presses are consumed as menu requests

The disabled-menu branch in `_maybeAnnotationMenu` is unconditional once a page point resolves. It invokes the callback and returns `true` even when:

- there is no annotation under the press,
- there is no annotation clipboard content,
- and `onContextMenuRequested` is null.

That prevents `_onLongPressStart` from reaching `_clearSelection()`. I reproduced this too: select “Page”, then long-press empty page space with `contextMenuEnabled: false` and no callback; the stale `"Page"` selection remains. The host callback is also spuriously called for empty page space.

This branch should only consume the gesture when a real menu target exists and a handoff actually occurs; otherwise it should return `false` so normal selection clearing continues.

### 3. The callback payload is insufficient for `PdfReader`

`PdfContextMenuHost` only receives positions and its docs recommend re-running `editing.selectableAnnotationAt(...)`. That works when the host already owns a `PdfEditingController`, but `PdfReader` owns its session internally and does not expose that controller. A `PdfReader` host therefore cannot reliably tell whether the request targets:

- page text,
- a normal or locked annotation,
- a form widget,
- or empty page space.

That blocks the PR's advertised “different actions per annotation type” use case for the drop-in reader. I suggest replacing the positional callback with a `PdfContextMenuRequest` carrying the already-resolved target/selection (similar to `PdfAnnotationMenuRequest`), or exposing an equivalent viewer-controller hit-test API.

All current GitHub checks are green; the first two findings came from targeted diagnostic tests against the PR head.

PR #567, opened 2026-07-24
fix(text): stop word spacing ballooning substituted table digits

ben-milanko:
## Summary

Reported: a page of ARTC ESC-21-03 (the "Busbar Voltage Leak to Earth Measurements" tables, §3.5.1/§3.6.1) rendered its numeric table cells as **enormous, distorted digits** — the leakage-volts columns ballooned to several times their size and overlapped, while the resistance column and body text were fine.

Every font on that page is a **non-embedded** standard TrueType (Arial/Verdana/Tahoma, no `FontFile`), so the renderer substitutes a system/bundled font. `CanvasPdfDevice` lays the run out as one string and stretches it horizontally so its total advance matches `PdfTextRun.width`. That width folds in **word spacing (Tw)**, and these cells use a large Tw (~8 em) on a trailing/leading space to reach the next column:

```
/TT2 1 Tf  0.0034 Tc 8.0586 Tw  7.98 0 0 7.98 110.1 664.5203 Tm
[(-)6( 0.)6(05 )2459(0.)6(1 )-7496(140000 )]TJ
```

So a segment like `"05 "` reported `width ≈ 9.7 em` while its visible glyphs are ~1.1 em, and the digits were stretched ~8× to fill it. (The flutter_test **Ahem** font hides this by drawing every substituted glyph as a solid box — the bug only shows with real fonts registered.)

The fix reproduces `Tc`/`Tw` as real spacing in the substitute layout, and hands painting devices the visible-glyph geometry separately so edge whitespace opens a genuine gap instead of stretching glyphs:

- `PdfTextRun` gains `letterSpacing`/`wordSpacing` (Tc/Tw in em; Tw is 0 for composite/CID fonts), `visibleWidth` (advance to the last non-whitespace glyph) and `leadingSpace` (advance to the first). All default to the previous behaviour.
- The interpreter accumulates the two advances in its glyph loop; the per-glyph visibility test is a non-allocating `_isBlankText` (matches Dart's `String.trim()` set) since it runs on the hot text path.
- `CanvasPdfDevice` applies the spacing via `TextStyle.letterSpacing`/`wordSpacing`, and for edge-whitespace runs paints a trimmed copy translated right by `leadingSpace`, stretched only to the visible-core width — matching Flutter's leading/trailing-trimmed `TextPainter.width`.
- `width`/`transform`/`text` stay whitespace-inclusive, so **text extraction**'s inter-run gap geometry is unchanged.
- The render-command codec round-trips the new fields (format `v3`) for the worker/transcript paint path.

Justified paragraphs on the same pages now spread correctly as a bonus. After the fix the page matches the expected reference exactly (dash in the 120V column, `0.05`/`0.5`/… in 50V, aligned percentage and resistance columns).

## Affected packages

- [x] `pdf_graphics`
- [x] `dart_pdf_editor`
- [x] Documentation / CI / repository metadata only (dev-log)

## Change type

- [x] Bug fix
- [x] Rendering change

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze` (clean, whole repo)
- [x] `cd packages/pdf_graphics && fvm dart test` (906 pass, incl. new interpreter + codec cases)
- [x] `cd packages/pdf_document && fvm dart test` (800 pass)
- [x] `cd packages/dart_pdf_editor && fvm flutter test` (render/text suites: text_cache, per_glyph_text, render_trace_gate, render_worker*, render_record_replay, editing_text_style/font_render, hover_text_warm, smoke — all pass)
- [x] Ghent corpus test (20 pre-existing colour/overprint deviations, identical count on `main` — not caused by this change) and pdfjs pixel-compare (164 pass)
- [x] `tool/perf.sh gate` — 13 counters within 3%
- [x] Real PDF render smoke test (ARTC ESC-21-03, before/after)

The reproducing PDF is a public ARTC engineering standard (not bundled). Regression coverage is a programmatic geometry contract test in `interpreter_test.dart` (trailing-space, leading-space and plain-run cases with exact Helvetica-AFM advances) plus a codec round-trip case. Before/after: the leakage-volts digits go from ~8× oversized/overlapping to correctly sized and column-aligned.

## Compatibility checklist

## Notes for reviewers

- Runs with an ordinary (Tw-free) trailing space are now also trimmed for the stretch factor, so such text is a few percent tighter (i.e. more correct — a trailing space no longer widens the visible word); positioning of following runs is unchanged because it reads the full advance. pdfjs pixel-compare stays green.
- Codec `_formatVersion` bumped 2→3 (producer/consumer ship together; invalidates any older cached transcript, which is the intended effect).

---
_Generated by [Claude Code](https://claude.ai/code/session_01TBrTtkwNoUQVtLwrYg8cmo)_


ben-milanko:
CI on `3b75249` failed on **one** test — `dart_pdf_editor/test/strip_parity_test.dart` ("strip device parity vs canvas device"). This is not caused by this change, and I'm not fixing it here — details below.

**What it checks:** it renders each Ghent page two ways (strip-binned vs direct-canvas) and asserts ≥90% of pages match within a structural-diff gate. It's a pixel-parity gate, sensitive to antialiasing at strip seams; the pages that trip it are all colour/overprint/transparency fixtures (`GWG030_Gray_K_black_OP`, `GWG040_White_OP`, DeviceN/Transp/OptionalContent) — not text.

**Why it's not this PR:** this change only touches substituted-font text with word/char spacing, and it moves the strip and canvas paths *identically*, so it's parity-neutral by construction. Confirmed empirically — same commit of the test, run locally:

- this branch: `57 pages | relaxed-edge gate pass 54 (94.7%)` → **pass**
- clean `origin/main` (worktree, without this change): `57 pages | relaxed-edge gate pass 54 (94.7%)` → **pass**

Identical numbers with and without the change, both comfortably over the 90% gate. The CI failure is CI's renderer tipping ~2–3 of those colour pages just under the marginal gate — environment-sensitive flake, reproduces on the base branch, predates this PR.

I've deliberately not touched `strip_parity_test`'s tolerance: loosening an unrelated colour-parity gate from inside a text-rendering PR would be scope creep and would mask a real signal for someone else. A re-run should clear it. The latest commit (`fcf4d23`, the CI-regenerated web-worker bundle) has a fresh CI run queued in `action_required` (awaiting approval) — approving/re-running that is the path to green.

---

Issue #403, opened 2026-07-19
perf(editor): pointer-event costs — overlay cursor notifier + hover hit-test ordering


ben-milanko:
## Summary

Two pointer-event-frequency costs in the viewer: the editing overlay rebuilds its entire ~7000-line subtree per mouse-move to move a cursor dot/swatch, and the viewer's `_onHover` runs up to 4 hit-test scans per event, including a full synchronous text extraction on first hover over a page.

## Evidence

1. Overlay rebuilds: `packages/dart_pdf_editor/lib/src/editing/editing_overlay.dart:4114-4199` — `setState` per hover event (4142, 4164, 4170, 4177, 4188, 4196) while ink/eraser/count/stamp/signature tools are armed and during eyedropper picking (`_updatePickPreview`, 4117). Only the painter needs the position.
2. Hover scans: `packages/dart_pdf_editor/lib/src/pdf_viewer.dart:3532-3557` — every `PointerHoverEvent` runs `_annotationAt` (3544), `_annotationHitAt` again when `onAnnotationTap` is set (3545-3547), `_selectableAnnotationAt` (3548), and `_textPositionAt(tolerance: 8)` (3550). Each redoes `_pagePointAt` (2893-2916) — linear page scan + fresh `PdfPageGeometry` per probe. `_textPositionAt` -> `_pageText` (2857) lazily runs `PdfTextExtractor.extract` — a full synchronous interpretation — the first time the mouse crosses a page.
3. Per-build rescans: `pdf_viewer.dart:581-584`, `:4948-4953` — `_matchesOn(index)` filters all matches per page per build; `_textSelectionOn` (4953) recomputes the same selection quads `_selectionQuadsOn` just computed (4125, 4315).

## Impact

Continuous build-phase cost during every armed-tool hover (the moment the user is about to draw); a one-time multi-hundred-ms freeze per page on first hover; per-event CPU on dense pages; scroll stutter during active searches.

## Proposed fix

1. Hoist cursor/eyedropper position into a `ValueNotifier<Offset?>` consumed by a repaint-bound `CustomPainter`; leave the rest of the overlay out of the rebuild.
2. `_onHover`: compute `_pagePointAt` once per event and share; cheap-hit-first ordering with early exits; debounce the text-cursor probe and extract text asynchronously (see #396).
3. Group search matches by page once when results land; memoize selection quads keyed on (page, anchor, focus).

## Dependencies

[...]

Issue #396, opened 2026-07-19
perf(editor): default render worker + off-UI-thread text extraction


ben-milanko:
## Summary

Page interpretation runs synchronously on the UI thread whenever no `PdfRenderWorker` is attached (opt-in today), and text extraction for search/hover always runs synchronously. A single 100-420 ms CAD page blocks the frame outright; search extracts every page on the UI thread yielding only every 5 pages (~2 s uninterruptible worst case).

## Evidence

- `packages/dart_pdf_editor/lib/src/renderer.dart:100-144` + `pdf_page_view.dart:867-930` — workerless path falls through to UI-thread interpretation.
- `packages/dart_pdf_editor/lib/src/pdf_viewer.dart:2803-2831` — `_searchAllPages` calls `PdfTextExtractor.extract` per page on the UI thread; yields every 5th page only (2828).
- First hover on an uncached page also triggers full synchronous extraction: `pdf_viewer.dart:2857` (`_pageText` via `_textPositionAt` in `_onHover`, 3550).
- `PdfTextExtractor` is pure Dart (pdf_graphics is VM-safe) — directly isolate-compatible.

## Impact

Scrolling jank on every uncached page, zoom-settle stalls, thumbnail population on workerless platforms (web without the worker script), search input freeze on large/CAD documents, one-time multi-hundred-ms hover freeze per page.

## Proposed fix

1. Ship a default `PdfRenderWorker` (isolate on native, web worker on web) so interpretation is off-UI-thread unless the host opts out.
2. Add an `extractText(pageIndex)` job to the render worker (or `Isolate.run`) for search and hover extraction.
3. Short term: yield every page (not every 5) in `_searchAllPages` and gate the hover text probe until extraction completes.

## Dependencies

- Related to #394 (single-walk default) — land together or in either order; the worker jobs should use whichever interpretation path survives #394.

## Suggested order

Phase 2, ticket 5 of 16.

ben-milanko:
## Verification pass — CONFIRMED (one line number drifted)

All four workerless paths verified:

- **Web without worker script** — `render_worker_web.dart:74-84`: `startRenderWorker` returns `_WebRenderWorker.disabled()` when `pdfRenderWorkerScriptUrl == null` or the `Worker` constructor throws; a worker-level `onerror` calls `_fail()` (`:93-96`) making `isActive` false (`:167`). Hard stub also exists: `render_worker_stub.dart:17`.
[...]

PR #599, opened 2026-07-25
perf(editor): hoist the overlay's hover cursors onto a repaint layer (#403 part 1)

ben-milanko:
## Summary

Closes the last open part of #403.

With ink, the eraser, count, stamp or the signature tool armed — or the eyedropper picking — the editing overlay paints its own cursor, and the cursor's position was a **snapshot argument of `_EditingPreviewPainter`**. Moving it therefore meant `setState`: every `PointerHoverEvent` rebuilt `_EditingPageOverlayState`'s whole subtree (annotation widgets, selection chrome, the ~7000-line build method, and a fresh preview painter with ~50 recomputed arguments) to move a dot a few pixels — continuous build-phase cost through exactly the moment the user is about to draw.

The fix is a sibling of the pattern the file already had for the in-progress stroke (`_ActiveStrokePainter` + `_activeStrokeRepaint`):

- **`_HoverCursorPainter`** — a new painter on its own `RepaintBoundary`, `super(repaint: _state._cursorRepaint)`, reading the live `_penCursor` / `_eraserCursor` / `_countCursor` / `_stampPreview` / `_rotateCursor` / `_signaturePreview` fields straight off the state. It sits above the active-stroke layer, so the cursors are topmost among the painted layers (they used to be the tail of the preview painter, *below* the stroke).
- `_onHover`, the signature drag and `_extendActiveStroke` assign the field and bump `_cursorRepaint` instead of `setState`. The one hover outcome that still rebuilds is a change of **system cursor kind**, which lives on the `MouseRegion` and only changes at tool/zone boundaries, not per event.
- The eyedropper's swatch is a widget, so it can't join the painter — it subscribes to the same notifier through a `ValueListenableBuilder`, rebuilding the chip alone.
- `_EditingPreviewPainter` loses `eraserCursor`/`eraserRadius`/`penCursor`/`penOpacity`/`countPreview`/`stampPreview`/`rotateCursor`, and the signature preview leaves its `extraInk`. `_paintStampAfterimage` and the three stamp-template helpers become top-level functions taking the view `scale`, so both painters draw the identical mark.

No behaviour change is intended: same marks, same gates, one layer higher in z-order.

Parts 2 and 3 of #403 already landed (#442, #479); `pdf_viewer.dart` is untouched here.

## Affected packages

- [ ] `pdf_cos`
- [ ] `pdf_document`
- [ ] `pdf_graphics`
- [x] `dart_pdf_editor`
- [ ] `pdf_test_fixtures`
- [x] Example app *(perf-harness tooling under `app/tool/perf/` only — no app code)*
- [x] Documentation / CI / repository metadata only *(dev-log + perf README)*

## Change type

- [ ] Bug fix
- [ ] Feature
- [ ] Rendering change
- [ ] Editing behavior change
- [x] Performance change
- [ ] Refactor / cleanup
- [ ] Tests / fixtures only
- [ ] Documentation only

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze` — clean at repo root
- [x] `cd packages/dart_pdf_editor && fvm flutter test` — **1969 passed**
- [ ] Example app smoke test

If any relevant check was not run, explain why: the change is confined to `dart_pdf_editor`'s widget layer — it touches no COS/parse/render path, so the pure-Dart package suites and the Ghent/corpus render baselines have nothing to exercise. No render baseline could move: the same marks are drawn, only on a different layer.

### Performance

Two independent checks, because they answer different questions.

**Structural (widget tests).** `_EditingPreviewPainter`'s *identity* is the tell — a rebuild constructs a fresh one from new arguments, so the same instance still being mounted after a pointer move means no rebuild ran. `editing_cursor_test.dart` asserts that for ink / eraser / count / stamp / signature, plus (so the assertion can't quietly go vacuous) that a change the overlay really does rebuild for — setting the controller colour — *does* swap the painter out.

**Wall-clock (real Chrome).** The web harness gains a `hover` kind (`hover-ink`, `hover-eraser`): arm a tool, dispatch N synthetic `PointerHoverEvent`s at frame cadence over a dense vector sheet, report `hoverBuildMs*` from the `FrameTiming`s across the sweep. `tool/perf.sh webdiff HEAD hover-ink`, 3 iterations per side, 240 events over `plan-set-16p.pdf`, medians:

| metric | baseline | this branch | Δ |
|---|---|---|---|
| `hoverBuildMsP50` | 1.46 | 1.10 | **−24.6%** |
| `hoverBuildMsTotal` | 527 | 493 | −6.5% |
| `hoverBuildMsPerEvent` | 2.20 | 2.05 | −6.5% |
| `hoverBuildMsP95` | 4.96 | 5.08 | +2.5% |

Reading these honestly:

- `hoverBuildMsP50` is the clean result — every working-tree run measured 1.1 ms, every baseline run 1.4–1.7 ms, so the ranges don't overlap. A quarter off the *typical* build frame during an armed hover.
- The totals overlap run to run (WT 468/493/524 vs baseline 525/527/620). −6.5% is the right sign, but it's diluted by design: 240 frames over a 16-page plan set contain plenty of build work the overlay has nothing to do with, and the harness can't attribute build time per widget. The widget tests are what prove the overlay's own share went to zero.
- **The tool also prints `hoverBuildMsMax` 33.35 → 7.67 ("−77%"). That is not a result** — both sides threw a ~40 ms outlier in one of three runs (WT 7.7/7.6/40.9, baseline 33.4/40.8/7.4); the medians only separate because the outliers landed differently. Please don't quote it.
- The tool flags `buildMax` 71.3 → 78.5 as a regression. That's the *whole-run* max, and the run has only ~8 frames outside the hover window — it's a cold-open first-paint frame, which this change can't touch. The in-window max is `hoverBuildMsMax` above.

None — no document bytes change. The one visual difference is z-order: the painted cursors now sit above the in-progress stroke instead of below it (strictly better; the eraser ring was already documented as "topmost" within its old layer).

## Compatibility checklist

- [x] Parsers remain lenient on real-world input and writers remain strict on output. *(untouched)*
- [x] Raw PDF stream bytes stay lazy/raw until decoding is required. *(untouched)*

## Notes for reviewers

- **`shouldRepaint` is `true` on the new painter**, unlike `_ActiveStrokePainter`'s `false`. The stroke layer's notifier ticks on every mutation of its buffers, so it can own its repainting outright. The cursor layer also depends on inputs it does *not* observe — armed tool, pen colour/width, zoom (`_geometry`), `_chromeScale` — which change through ordinary rebuilds. A fresh painter instance means the overlay rebuilt for one of them; repainting a few circles and an arc then is cheaper than tracking which input moved.
- **The paint gates are the debug getters.** `_state._eraserCursor` outlives the eraser (nothing clears it when the tool is put away); the `eraserCursor` getter — what `paint()` reads — is what's actually drawn, so the tests assert on the getters rather than the raw fields.
- **Deliberately left on `setState`:** `_polyHover` (polygon/cloud rubber band) and the erase-drag preview. Both change what the *page* preview draws — `dragPath`, the sliced/faded stroke sets — not just a cursor, so they belong to the preview painter.
- The remaining #403 follow-up, async/off-thread hover text extraction, is still gated behind #396.
- Reviewers who want the numbers themselves: `tool/perf.sh webdiff main hover-ink` (or `hover-eraser`) now works out of the box.

---
_Generated by [Claude Code](https://claude.ai/code/session_01P2bAjHQowJKKgqVeCvZUAb)_

### DIFF
--- a/app/tool/perf/README.md
+++ b/app/tool/perf/README.md
@@ -25,7 +25,7 @@ measures.
 Declared in [`scenarios.json`](scenarios.json) — a name → `{kind, pdf, params}`
 registry. `pdf` is a repo-relative file from the CC0 `test_corpora/dartpdf`
 corpus, so **any checkout (and any A/B worktree) can run every scenario** with
-no magic local file. Four workload kinds ship:
+no magic local file. Five workload kinds ship:
 
 | kind | what it measures | headline metrics |
 |---|---|---|
@@ -33,9 +33,11 @@ no magic local file. Four workload kinds ship:
 | `open` | cold-open profile: bytes → `PdfDocument.open` → `pageCount` → first painted content on a target page | `openBytesMs`, `openDocMs`, `openPageCountMs`, `openFirstContentMs` |
 | `search` | full-document text search latency + hit count (best-of-N) | `searchMs`, `searchMatches` |
 | `edit` | apply a batch of annotations through the real `PdfEditingController`: incremental-save + appearance-gen cost | `editApplyMs`, `editApplyMsPerOp`, `editRevisions`, `editBufferGrowthKb` |
+| `hover` | mouse-move over a page with an editing tool armed: what following the painted cursor costs per pointer event | `hoverBuildMsTotal`, `hoverBuildMsPerEvent`, `hoverBuildMsP50/P95/Max`, `hoverFrames` |
 
 Stock scenarios: `scroll-plan`, `scroll-scan`, `scroll-diagram`, `open-plan`,
-`open-text`, `search-text`, `edit-annotate` (default `scroll-plan`).
+`open-text`, `search-text`, `edit-annotate`, `hover-ink`, `hover-eraser`
+(default `scroll-plan`).
 
 ### Adding a scenario going forward
 
@@ -123,6 +125,7 @@ searchMs                52.30       38.10       -27.2%    ✓ faster
 | `PERF_TARGET_PAGE` | scenario | open/scroll target page |
 | `PERF_QUERY` / `PERF_REPEAT` | scenario | search needle / best-of-N |
 | `PERF_OPS` | scenario | edit: annotations to apply |
+| `PERF_EVENTS` / `PERF_TOOL` | scenario | hover: pointer events to dispatch / the armed tool |
 | `PERF_IMAGE_CACHE_MB` | `0` | decoded-image cache budget, MB |
 | `PERF_HEADLESS` | `true` | `false` for a visible window |
 | `PERF_TIMEOUT` | `300` | overall budget, seconds |
--- a/app/tool/perf/perf_harness.dart
+++ b/app/tool/perf/perf_harness.dart
@@ -535,13 +535,17 @@ class _PerfHarnessAppState extends State<_PerfHarnessApp> {
     final editing = _editing!;
     // Let the first page paint before the pointer starts moving over it.
     await Future<void>.delayed(const Duration(milliseconds: 1200));
+    // A typo'd ?tool= must not quietly arm nothing and then report numbers for
+    // a workload that never happened - only the explicit 'none' disarms.
     editing.tool = switch (_hoverToolName) {
       'ink' => PdfEditTool.ink,
       'eraser' => PdfEditTool.eraser,
       'count' => PdfEditTool.count,
       'stamp' => PdfEditTool.stamp,
       'select' => PdfEditTool.select,
-      _ => null,
+      'none' => null,
+      _ => throw StateError('unknown ?tool=$_hoverToolName '
+          '(ink|eraser|count|stamp|select|none)'),
     };
     await Future<void>.delayed(const Duration(milliseconds: 300));
 
--- a/app/tool/perf/perf_harness.dart
+++ b/app/tool/perf/perf_harness.dart
@@ -22,6 +22,9 @@
 //   edit    - apply a batch of annotations (highlight/rectangle/ink) through
 //             the real PdfEditingController; incremental-save + appearance-gen
 //             cost, revision count, and session-buffer growth.
+//   hover   - mouse-move over the page with an editing tool armed: synthetic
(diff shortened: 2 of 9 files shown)


ben-milanko:
Self-review follow-up landed in `bb89c45`, since the description above predates it (and a squash-merge makes this the record).

**Two stroke-start sites were missing a `_cursorRepaint` tick.** `_onPointerDown`'s raw/stylus path and `_panStart`'s ink case assign `_penCursor` and then tick only `_activeStrokeRepaint`. That was correct while the pen dot was painted by `_ActiveStrokePainter`; it's stale now the dot lives on the cursor layer. Both now bump the cursor layer too, and `_ActiveStrokePainter`'s doc states the invariant so it isn't quietly reintroduced.

**It is hygiene, not a user-visible fix — worth being explicit about**, because the natural reaction to "no test fails without it" is to delete the bumps:

- The field is assigned correctly either way; only the repaint *scheduling* was missing.
- It isn't observable today. Both sites immediately seed a one-point `_activeStroke` and tick the stroke layer, and `_paintInkStrokes` draws a single point as a dot — same position, same colour, same frame as the pen cursor would have. The dot you see is the stroke's.
- It becomes observable as soon as either coincidence stops holding (stroke layer skipping single-point strokes, the cursor gaining a distinct look, a stroke start that doesn't seed a point).

Two attempts at a failing test each established that rather than the bug: a getter-based assertion can't observe repaint scheduling at all, and a pixel assertion at the press point sees the stroke's own dot. So the test that landed is named for what it actually pins — that the raw-pointer path positions the pen cursor with no hover first — and `doc/dev-log/2026-07-25-overlay-cursor-repaint-layer-403.md` records why no failing test exists.

Also in that commit: the `hover` perf scenario now throws on an unknown `?tool=` instead of silently arming nothing and reporting `hoverBuildMs*` for a workload that never ran.

Suite is **1975 passed** on `bb89c45` (the 1969 in the description is pre-follow-up); `dart analyze` clean at root.

---

Issue #603, opened 2026-07-25
perf(editor): the thumbnail warmer spends ~6.7 s of main-thread replay in a 20 s scroll


ben-milanko:
## Measured on device

Web release, app 3.0.0+20 build with #601, 62-page / 24 MB book, 3 render workers. A ~20.8 s capture (perf 31421 → 52184) of scrolling pages 27→41.

Totalling every `thumbnail page=N … worker record=` line in that window:

- **~17.0 s of record wall-clock** (queue included)
- **~6.7 s of main-thread `replay`**

The replay figure is the problem. `rasterizeThumbnail` builds its `ui.Picture` via `PdfPageRenderer.pictureFromCommands`, which runs on the platform thread — the same one the visible page's build needs, and the one that cannot move to an isolate.

```
thumbnail page=47 warm px=256 worker record=394.5ms replay=859.6ms raster=10.2ms
thumbnail page=31 warm px=256 worker record=535.9ms replay=510.9ms raster=26.7ms
thumbnail page=32 warm px=256 worker record=901.3ms replay=482.2ms raster=11.7ms
thumbnail page=33 tile px=256 worker record=359.0ms replay=444.4ms raster=14.3ms
thumbnail page=42 warm px=256 worker record=479.8ms replay=376.8ms raster=20.0ms
```

**859 ms of main-thread work to draw a 256-pixel-wide thumbnail.** For scale, the actual page builds in the same capture are `build=44–472ms` — the thumbnails cost more than the pages they preview.

## Why

The warmer replays the **full** command record and then rasterizes it down to 256px. Page 40 records 1787 commands, page 34 1414, page 35 1588 — all replayed in full for a 256px tile.

`rasterizeThumbnail` does pass `imagePixelRatio: pixelWidth / size.width` (≈0.42 at px=256, `editing_thumbnails.dart`), so *decoded* pixels are capped. But that cap does nothing about command volume, and nothing about images the worker's codec declines — a declined image ships its **source stream** instead, which no pixel ratio shrinks:

```
page=49 … 9955392B … declined=58(decodeNull:7,notDct:51)
page=51 … 10569471B … declined=57(decodeNull:7,notDct:50)
```

Those are ~10 MB records produced to fill a 256px thumbnail.

## It runs during active scrolling

[...]

PR #607, opened 2026-07-25
perf(scheduler): stop granting a page a second concurrent render (#451)

ben-milanko:
Item 3 of the [#601 write-up](https://github.com/ben-milanko/dart-pdf/blob/main/doc/dev-log/2026-07-25-record-glyph-outline-dedup-451.md) listed two untracked scheduler problems the device trace exposed. This is the second — "every focus page is granted exactly twice" — chased to a mechanism, fixed, and pinned.

**Read the measurement section before merging: the fix is correct and exercised, but no local scenario shows a win.**

## The race

`request` documents "the page is interpreted at most once" and dedupes by token, but only against `_pending` — and `_drain` *removes* the request before invoking it:

```dart
final next = _pending.removeAt(pick);
next.render();                       // returns at its first await
await SchedulerBinding.instance.endOfFrame;
```

`render` was typed `VoidCallback`, and Dart coerces the `Future<void> Function()` both call sites pass — so the future was dropped on the floor. On the worker path the callback returns in microseconds and the recording lands hundreds of ms later. A rebuild inside that window finds an empty `_pending`, enqueues, and the next `endOfFrame` grants the same page a **second concurrent pass**.

Both passes record the page in full. `_renderNow` opens with `beginFull()`, so the newer pass bumps the generation and the older one's result is discarded by its own `_superseded` guard when it lands — a whole wasted record. `pdf_page_view.dart:1725` already described the interleaving as a known hazard and defended the *state* against it; nothing stopped the duplicate work.

The device trace's smallest same-page gap was 3 ms — exactly one `endOfFrame` when a frame is already in flight.

## The fix

The scheduler tracks the in-flight window. A re-request for a token whose render has not settled is stored rather than queued; `_settle` (hung off the returned future, with `onError` so an async throw cannot wedge a page out of the queue) grants it once the running pass finishes. `cancel` drops a queued re-request so a recycled page is not granted again.

Held, **not dropped** — the rebuild may carry a new scale or revision. By the time it runs the page has its picture, so it takes `_renderNow`'s `cached != null` branch.

Frame pacing is deliberately unchanged: `_drain` still awaits only `endOfFrame`, never the render. Awaiting renders would serialize every page behind the slowest worker round trip and defeat three workers. A test pins that.

`VoidCallback` → `FutureOr<void> Function()`; both call sites already passed async functions. `_settle` pings `activity` so #603's thumbnail warm stands down for a re-queue as it would for a fresh request.

## Measurement — and why it shows nothing

Interleaved web A/B, `scroll-scan` (the image/cache-heaviest scenario), 3 iterations vs the pre-fix commit: every metric inside noise, and the grant log is **byte-identical modulo a startup offset** — 42 grants, 4 per page, `worker=31 recorded=0` both sides. (`jankCount` 1 → 2 tripped the gate; it counts log lines, and the run before went the other way.)

To find out whether the race fires at all, the coalesce now logs `scheduler defer page=N (render in flight)`:

| scenario | grants | defers |
|---|---|---|
| scroll-scan | 42 | 10 |
| scroll-plan | 25 | 5 |
| scroll-diagram | 6 | 0 |
| scroll-cad-labels | 8 | 0 |
| scroll-hatch | 8 | 0 |
| open-diagram-progressive | 2 | 0 |
| edit-annotate | 1 | 0 |

It fires — but what it catches locally is benign:

```
[perf 2359] scheduler grant page=2
[perf 2651] webworker result kind=record page=2 5589896B → worker
[perf 2651] scheduler defer page=2 (render in flight)
[perf 2857] scheduler grant page=2
[perf 2859] interpret page=2 path=worker FIRST interpret=1.5ms wait=0.0ms
```

The re-request arrives *as the record lands*, and the deferred grant replays the record it already has in 1.5 ms. One record, not two. Serializing it costs nothing and saves nothing — which is why `worker=` does not move.

The wasteful shape is in the device trace and is not reproduced locally:

```
page=28  worker=1895ms  ser=1608ms
page=28  worker=1723ms  ser=1505ms   <- same page, again
```

Local records are ~450 ms over a 12-page corpus; the device ran a 62-page book on three workers with records an order of magnitude slower, which is what widens the window enough for a second *recording* pass to start. **Confirming the win needs a device trace on this build** — count `scheduler defer` lines and check whether any page still produces two large records.

## Adjacent, not chased

The same log shows page 1's 41277 B vector-only record produced at `[perf 169]` and again at `[perf 2924]` — an identical re-record the record cache should have served. Different problem.

## Testing

Four new tests, each confirmed to fail against the specific mutation it targets: in-flight dedup removed, `cancel` leaving the re-request, `onError` dropped, `_settle` discarding instead of granting. dart_pdf_editor 1989, analyzer clean.

### DIFF
--- a/doc/dev-log/2026-07-26-scheduler-concurrent-grant-451.md
+++ b/doc/dev-log/2026-07-26-scheduler-concurrent-grant-451.md
@@ -1,0 +1,127 @@
+# 2026-07-26 — #451: the scheduler could grant one page two concurrent renders
+
+Item 3 of the [#601 write-up](2026-07-25-record-glyph-outline-dedup-451.md)
+listed two untracked scheduler problems the device trace exposed. This is
+the second one — "every focus page is granted exactly twice" — chased to a
+mechanism, fixed, and pinned. **The fix is correct and exercised, but no
+local scenario shows a measurable win.** That caveat is the point of this
+note.
+
+## The race
+
+`PdfPageRenderScheduler.request` documents "the page is interpreted at most
+once", and deduplicates by token. But it only compares against `_pending`,
+and `_drain` *removes* the request before invoking it:
+
+```dart
+final next = _pending.removeAt(pick);
+next.render();                       // returns at its first await
+await SchedulerBinding.instance.endOfFrame;
+```
+
+`render` was typed `VoidCallback`, and Dart happily coerces the
+`Future<void> Function()` both call sites pass — so the future was dropped
+on the floor. On the worker path the callback returns in microseconds and
+the recording lands hundreds of milliseconds later. A rebuild inside that
+window finds an empty `_pending`, enqueues, and the next `endOfFrame`
+grants the same page a **second concurrent pass**.
+
+Both passes record the page in full. `_renderNow` opens with
+`_renderSession.beginFull()`, so the newer pass bumps the generation and the
+older one's result is discarded by its own `_superseded` guard when it
+lands — a complete wasted record. `pdf_page_view.dart:1725` already
+described the interleaving as a known hazard and defended the *state*
+against it; nothing stopped the duplicate work itself.
+
+The device trace's smallest same-page gap was 3 ms, which is exactly one
+`endOfFrame` when a frame is already in flight.
+
+## The fix
+
+The scheduler now tracks the in-flight window: `request` for a token whose
+render has not settled stores the re-request instead of queueing it, and
+`_settle` (hung off the returned future, with `onError` so an async throw
+cannot wedge a page out of the queue) grants it once the running pass
+finishes. `cancel` drops a queued re-request so a recycled page is not
+granted again when its abandoned render lands.
+
+The re-request is **held, not dropped** — the rebuild that asked may carry a
+new scale or revision. By the time it runs the page has its picture, so it
+takes `_renderNow`'s `cached != null` branch and re-rasters rather than
+recording from scratch.
+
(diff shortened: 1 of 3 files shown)


ben-milanko:
## Device trace: the race is closed

Confirmed on device (app 3.0.0+21, web release, 3 workers). Seven `scheduler defer` lines fired, and every same-page re-grant now sits behind the first pass:

| page | first grant | defer | second grant | gap |
|---|---|---|---|---|
| 9 | `[3308]` | `[3309]` | `[4117]` | 809 ms |
| 21 | `[8305]` | `[8435]` | `[8479]` | 174 ms |
| 25 | `[10616]` | `[10618]` | `[10730]` | 114 ms |
| 31 | `[13297]` | `[13312]` | `[13480]` | 183 ms |

Pre-fix, the two grants for a page landed 3 ms apart (`[33439]`/`[33442]`) — two concurrent recordings, one discarded. Pages 9 and 25 are the clearest: the re-request arrives **1–2 ms** after the grant, exactly the window that used to start a second concurrent pass.

Every same-page grant pair in the trace has a `defer` between it; none is closer than 114 ms. No duplicate concurrent grant remains.

Honest limit: this shows the race is closed, not that the run got faster — that needs a matched before/after over one scroll. The multi-record pages (1, 9, 29, 30, 31) now produce records for *distinct* purposes (vector-first, full, prerender warm, thumbnail) separated in time, rather than two identical full records back to back.

Issue #502, opened 2026-07-22
render: faithful subtractive overprint needs a CMYK/spot colorant buffer (3 GWG030 patches)


ben-milanko:
> **Rescoped 2026-07-25.** Overprint *is* implemented — this ticket originally read "the renderer does not implement overprint at all", which is no longer true. Parsing (#506) and the RGB `darken` approximation (#515, ceiling pinned by #518) both landed. What remains is the last, architectural step: a colorant buffer. The original framing is preserved in the history below.

## What's left

Three GWG030 patches still show their self-grading fail-marker "X", and **no RGB compositor can clear them**:

| patch | case |
|---|---|
| d | 50% K over CMYK, OPM 0 |
| e | 50% gray over CMYK, either mode |
| k | 50% gray over CMYK, either mode |

In each, a neutral ink must knock the DeviceCMYK backdrop's process colorants out to grey — while a **spot** backdrop of the *same RGB colour* must survive untouched. Those two cases are indistinguishable once everything is RGB, so the current `BlendMode.darken` approximation cannot separate them. It is the RGB optimum: empirically no OPM-keyed blend does better (`doc/dev-log/2026-07-23-overprint-rgb-ceiling.md`).

**The fix is a CMYK/spot colorant buffer** — composite affected draws in ink space with OPM-0/1 semantics, then convert. It interacts with Separation/DeviceN tint transforms and transparency groups, which is why it was split out rather than attempted inline.

## Current state (what already works)

- `/OP`, `/op`, `/OPM` parse into the graphics state and reach devices via `PdfDevice.setOverprint` (#506).
- `CanvasPdfDevice` composites an overprinting fill/stroke/gradient/mesh with `BlendMode.darken` (per-channel min) while the blend mode is Normal (#515). `darken` is a no-op over white, so the ~40 Ghent pages that set `op`/`OP` defensively over the page background are unaffected. `multiply` was tried and rejected — it double-darkens ink over ink.
- `StripBinningDevice` routes overprinting draws to its delegate, holding strip/canvas parity.
- `CanvasPdfDevice.debugOverprintCompositing` is a kill switch (it was the tool that separated overprint from #505's transparency-group change while root-causing #550).
- Markers now flatten on **every "over spot" patch** and on the "over CMYK" patches whose neutral ink only darkens (K under OPM 1, separation black).
- Per-patch behaviour — which markers flatten, which remain — is pinned platform-independently by `overprint_render_test.dart`, so this can't silently regress.

[...]

PR #608, opened 2026-07-25
feat(render): faithful image overprint via per-sample colorant readings (#604)

### LINKED ISSUE #604: render: image overprint (colorant reading for decoded rasters)
Follow-up to #502 / #602. The CMYK/spot colorant buffer made **vector** overprint faithful, but **image** overprint is still unimplemented: a decoded raster carries no colorant reading, so an image neither overprints what's under it nor accepts overprint from ink drawn over it.

## What still fails

Left over from #502's acceptance clause "as fidelity improves, GWG190/191/192 should follow":

- **GWG190/191/192 DeviceN Overprint** — the two *vector* patches on each page are now solid/correct (#602); the two *image* patches still show their self-grading "X" marker. These pages remain in `ghent_render_test.dart`'s `_knownBaselineDeviations`.
- **GWG031 Gray Image Overprint** — unchanged; image overprint out of scope for the colorant buffer.

## Why it's separate

`overprint_compositor.dart` records the colorants each *vector* draw leaves behind and composites an overprinting draw against them. An image draw has no such reading — the buffer would need each image resolved to process/spot colorants (its colour space + samples), which interacts with image color spaces (DeviceN/Separation/ICC images), `/Decode`, and masking. That's a materially larger change than the vector path.

## Acceptance

Remove GWG190/191/192 (and, if reachable, GWG031) from `_knownBaselineDeviations` and pixel-enforce, with per-patch coverage extended in `overprint_render_test.dart`.

See `doc/dev-log/2026-07-25-overprint-colorant-buffer.md` for the vector-side design and the "still out of scope: image overprint" note.

ben-milanko:
## Summary

Closes #604. The CMYK/spot colorant buffer (#502) made **vector** overprint subtractive by recording the device colorants each draw leaves behind. An **image** draw had no such reading — decoded pixels are sRGB, which has no colorant decomposition — so a raster neither overprinted what was under it nor accepted overprint from ink drawn over it.

This supplies the reading where it actually exists: the image's own *samples*, in its own colour space, before they become pixels.

- **`PdfColorSpace.inkColorants` for `/Indexed`** — reads the *base* space through the selected palette entry (§8.6.6.3); null for an RGB-ish palette, like every other space with no reading. GWG031's gray raster is `/Indexed` over DeviceCMYK, so without this it had no reading at all.
- **`pdf_graphics/lib/src/image_colorants.dart`** (new) — `pdfImageColorants` answers "what colorants does this raster carry"; `pdfImageOverprintStream` composites every sample over a backdrop (`PdfInkColorants.over`) and returns a **substitute image stream** whose colours are already the composite.
- **`PdfOverprintCompositor.image`** — the policy: rasterize the image's quad, read the backdrop under it, offer it, record what the raster leaves behind. Its sibling **`.stencil`** covers the one image kind that needs no substitute: an `/ImageMask` paints the *fill* colour through its own alpha, so only that colour needs resolving.

A substitute XObject rather than a new device call, a flag on `PdfImageRequest`, or a per-pixel hook: an image XObject is a thing every path already knows how to draw, so the canvas device, strip device, recorded command buffer, render worker, codec and decoded-image cache need **no change at all** — the interpreter simply draws a different stream.

Two rules carried over from the vector path: the exact-answer rule (a composite equal to the backdrop reuses the backdrop's own rendered pixels, one equal to the ink reuses the ink's, so an overprint that changed nothing is invisible rather than a shade off), and bare paper as the identity (overprinting onto no colorant leaves every sample as it was, so `pdfImageOverprintStream` returns null and nothing is built — the case that keeps pages setting `/op` defensively over white free).

API impact: `PdfImageColorants`, `pdfImageColorants`, `pdfImageOverprintStream` and `PdfOverprintCompositor.image`/`.stencil`/`.spotEquivalents` are additive; no existing signature changes outside the interpreter's private helpers.

## Affected packages

- [ ] `pdf_cos`
- [ ] `pdf_document`
- [x] `pdf_graphics`
- [x] `dart_pdf_editor` (tests + Ghent baselines only)
- [ ] `pdf_test_fixtures`
- [ ] Example app

## Change type

- [ ] Bug fix
- [x] Feature
- [x] Rendering change
- [ ] Editing behavior change
- [ ] Performance change
- [ ] Refactor / cleanup
- [ ] Tests / fixtures only
- [ ] Documentation only

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze` (`--fatal-infos`, clean)
- [x] `cd packages/pdf_graphics && fvm dart test` (982 passing)
- [x] `cd packages/dart_pdf_editor && fvm flutter test` (1984 passing)
- [x] Ghent corpus test and baseline update — 44 pixel-enforced + 10 tolerated, all passing
- [ ] Real PDF corpus parse/render smoke test — the `corpus/` set is git-ignored and not present in this environment
- [ ] Example app smoke test — no UI surface changed

`pdf_cos` and `pdf_document` are untouched by this change, so their suites were not re-run.

**Performance.** `tool/perf.sh gate` (the deterministic half) is **re-baselined**: GWG010 and GWG020 decode more streams now — the images' own samples, plus the tint-transform streams of the colour spaces that give them a colorant reading. Both are pages that declare overprint *and* draw colorant-bearing images, and on both the reads are used (a substitute is built), so this is the feature's own cost rather than speculative work; the other ten gate inputs are unmoved. See the comment thread for the trace.

By construction the work is bounded: a raster whose colour space has no colorant reading (DeviceRGB/ICCBased/Lab — most image content) is refused *before* `decodeStreamData` is called, so nothing is decompressed to find out; a colorant-bearing raster pays one uniformity scan per stream per document, memoised; and a substitute is built only for a raster that actually overprints onto something other than paper. The wall-clock `tool/perf.sh diff` A/B over `ghent-suite-open` is still running — numbers to follow in a comment.

Fixtures are the checked-in Ghent suite. Six pages change, all overprint pages:

| page | before → after |
|---|---|
| GWG190/191/192 DeviceN Overprint | 2 of 4 patches solid → **all 4** (out of `_knownBaselineDeviations`) |
| GWG031 Gray Image Overprint | the "Wrong" white box around the drop shadow → the **"Correct"** green |
| GWG020 Spot to CMYK Overprint | 4 of 10 self-grading patches passing → **8 of 10** |
| GWG010 CMYK Overprint Test | its OPM-1 "mask" patch now shows the backdrop through |

GWG190/191/192's own legends state the grading verbatim: "a + b must be rendered to solid Black (100C100K) rectangles… a faint X means that OP is not honored (or image was converted to DeviceCMYK for b)"; "c + d must be rendered to solid Cyan rectangles".

## Compatibility checklist

- [x] Raw PDF stream bytes stay lazy/raw until decoding is required — the sample reader only decodes a stream once the colour space has already said it has a colorant reading.

## Notes for reviewers

**Three intentional baseline changes worth a look.**

1. **GWG190/191/192** leave `_knownBaselineDeviations` and become pixel-enforced (41 → 44 enforced pages).

2. **GWG020 was re-seeded and stays tolerated.** Six more of its ten patches now pass: the two image and two shading patches (a uniform raster is now a *backdrop* the buffer can composite against — the other half of this issue, which the acceptance criteria didn't have a fixture for) and the two `/ImageMask` "mask" patches (from `.stencil`). Its two **font** patches still show their marker and will until text is modelled by its outlines rather than its em box — #502's deliberate cost trade — so the page stays in the set.

3. **GWG010 moved and was re-seeded.** It is a visual-comparison page with no self-grading legend. Its "mask" patch paints a rich-black X (`.5 .5 .5 .5 k`), then draws an `/Indexed`-over-DeviceCMYK 50%-magenta raster clipped to that X with `/op true`. Under OPM 1 the raster's three zero components leave the backdrop's C, Y and K standing, so the X shows through; under OPM 0 it writes all four and knocks the X out to flat pink. That contrast between the two rows is the page's subject, and the rule producing it is the same one GWG031 grades explicitly ("Overprinting must be honored for the output") and GWG011 pins for vector paint ("if an X appears the Overprint Mode is not respected"). **If you read that patch differently, this is the line to push back on.**

**A subtle interaction that cost more thought than the compositing.** The decoded-image cache keys XObject images by *stream identity*, and one render path (`renderPictureWithPlan`, plus the strips harness) discovers images with a separate scan-only walk before painting. So (a) substitutes are memoised per `(stream, backdrop, mode)` on an `Expando`, and (b) `_beginOverprint` now clears `_scanImages` for any page that opens a buffer — a page that resolves overprint is walked in full even to *collect* its images. Without either, the paint walk would ask the cache for a stream the collect walk never decoded and the image would silently fail to appear (a blank panel, not a wrong colour). `overprint_render_test.dart` has a dedicated guard rendering GWG031 through the two-walk path, because nothing else would catch that.

**Known limitation / follow-up.** One substitute composites against one backdrop, so a raster straddling two of them still declines. GWG010's neighbouring "image" patch is exactly that shape — the same test with the raster covering the whole square instead of just the X, so it sits on the pink square *and* the black X — and stays a flat knockout. Making it right means a per-pixel backdrop, i.e. carrying the buffer's own raster into the substitute rather than a single vector. That is the last piece of image overprint still missing.

Full design notes, including every decline and why it sits where it does: `doc/dev-log/2026-07-25-image-overprint-colorants.md`.

https://claude.ai/code/session_019ToHYCy9X3eh9vqg8koy3B


ben-milanko:
CI's `test` job was red on the deterministic counter gate. Fixed in fc18d1d, along with a second thing the investigation turned up.

### 1. The counter gate — re-baselined, deliberately

```
ghent:1-CMYK/GWG010_CMYK_OP_x3.pdf   bytesDecodedFlate: 20752 -> 29777
ghent:2-SPOT/GWG020_CMYKSpot_OP_x1a.pdf  objectsLoaded: 41 -> 47
                                       streamsDecoded:  7 -> 11
                                    bytesDecodedFlate: 11810 -> 26280
```

I traced each one before accepting it. Both pages declare overprint *and* draw colorant-bearing images, so the interpreter now decodes those images' samples to read their colorants, plus the tint-transform streams of the Separation/DeviceN spaces that give them the reading. On both pages the reads are **used** — a substitute raster is actually built — so this is the feature's own cost rather than speculative work. The other ten gate inputs are unmoved (`counter gate OK: 12 inputs, 13 counters within 3%`).

Worth stating explicitly since it isn't obvious from the counter names: a raster whose colour space has no colorant reading is refused *before* `decodeStreamData` is called, so DeviceRGB/ICCBased/Lab images — most image content — contribute nothing to this.

### 2. GWG020 went from 4/10 to 8/10 self-grading patches

Chasing the GWG020 counter shift made me look at what the page actually renders now, and six of its ten patches had cleared:

| | before | after |
|---|---|---|
| font (a, f) | ✗ | ✗ |
| vector (b, g) | ✗ | ✓ |
| image (c, h) | ✗ | ✓ |
| mask (d, i) | ✗ | ✓ |
| shading (e, j) | ✗ | ✓ |

The vector and shading patches came free: they sit on an image, and a uniform raster is now a *backdrop* the buffer can composite against — the other half of this issue, which I hadn't had a fixture for.

The **mask** patches needed one more piece, so I added it. An `/ImageMask` carries no colour of its own — it paints the *fill* colour through its own alpha (§8.9.6.2) — so its ink is ordinary vector ink with an ordinary colorant reading, and the only thing to resolve is the colour on the request; no substitute raster is involved. What the buffer can't model is a stencil's **coverage** (the mask, not the quad), so `PdfOverprintCompositor.stencil` resolves only when the backdrop is a single vector across the whole quad — where every pixel the mask does paint composites to the same colour, making "paint that colour through the mask" exact — and records the quad as unknown either way, since the backdrop survives wherever the mask is clear.

The two **font** patches still show their marker and always will under the current design: the buffer marks a text run by its em box rather than its glyph outlines, which is #502's deliberate cost trade (rasterizing outlines measured as most of the buffer's whole cost). GWG020 therefore stays in `_knownBaselineDeviations`; its baseline is re-seeded, and `overprint_render_test.dart` now pins its four image and stencil patches per-patch so the improvement can't silently regress behind that tolerance.

### Re-validated after both changes

- `dart analyze --fatal-infos` — clean
- `tool/perf.sh gate` — `12 inputs, 13 counters within 3%`
- `packages/pdf_graphics` — 978 passing (25 in the new `image_colorants_test.dart`, including four for the stencil path)
- `packages/dart_pdf_editor` — 1984 passing; the Ghent suite is 44 pixel-enforced + 10 tolerated, all passing
- `overprint_render_test.dart` — 6 tests, now covering GWG030 / GWG190 / GWG191 / GWG192 / GWG031 / GWG020

### Still outstanding

The `tool/perf.sh diff HEAD ghent-suite-open` A/B — my sandbox was restarted mid-run, so I've relaunched it and will post the wall-clock numbers here when it lands. Nothing about the shape of the change depends on the answer; the gate above is the deterministic half.

---

PR #609, opened 2026-07-25
perf(record): reuse a page's image decodes across its records (#451)

ben-milanko:
Follow-up to #607, whose device trace produced the numbers below.

That trace showed the scheduler race closed and left one cost standing above everything else — **≈6.9 s of `serialize=` across seven records, every one of them `cmyk:1`**:

```
[1438]  p2  serialize=908ms   cmyk:1
[2775]  p2  serialize=882ms   cmyk:1
[4031]  p2  serialize=910ms   cmyk:1
[4034]  p3  serialize=950ms   cmyk:1,notDct:2
[5556]  p3  serialize=918ms   cmyk:1,notDct:2
[15138] p29 serialize=1174ms  cmyk:1,notDct:1
[17412] p29 serialize=1148ms  cmyk:1,notDct:1
```

One `DeviceCMYK` JPEG per page the browser codec declines, decoded in pure Dart — and **re-paid on every record of that page**. A worker records the same page several times in one scroll (vector-first, full, prerender warm, thumbnail tile). Page 2 pays it three times; that repetition is ~4.6 s of the 6.9 s.

## The other lever died on contact — worth recording

The obvious first move was "cap the declined decode to the record's pixel budget": a 128 px thumbnail has no business decoding a 2 Mpx CMYK image. New `tool/bench_image_scale_cost.dart` decodes each image three ways:

```
  fullMs   pageMs   tileMs  tile/full  native      page        tile
   186.3    184.5    188.5       1.01  1280x1655   1280x1655   257x333   p2  CMYK SMask DCT
   191.2    186.0    188.2       0.98  1280x1655   1280x1655   257x333   p3  CMYK SMask DCT
   201.5    206.9    204.5       1.01  1241x1621   1241x1621   250x326   p29 CMYK SMask DCT
     7.1      5.0      2.6       0.37  1241x1621   1241x1621   250x326   p29 Gray Flate
```

**The thumbnail already asks for the small size and it costs the same.** #603's budget reaches the codec correctly; the *decoder* ignores the target for DCTDecode — it decodes the whole JPEG and downsamples. The Flate row at `0.37` shows that path does honour it. DCT has no scaled path the way JPX has `_decodeJpxScaled`.

Splitting further: JPEG entropy+IDCT inside `package:image` is only **~31%**; the other ~69% is our CMYK→RGBA/SMask/ICC pass at full resolution. A scaled DCT decode is therefore worth much more than 31% — but needs a reduced-IDCT decode `JpegData` doesn't expose. Left as follow-up.

## What this does

`PdfImageDecodeCache`: a bounded LRU threaded through `serializeCommands` as an optional `imageCache:`, owned one-per-open-document by both worker backends.

Deliberately **exact-match** — same stream, same requested target size — and it never downsamples a larger entry to serve a smaller request, because for formats that *do* have a scaled decode path that would substitute different pixels for the ones the decoder would have produced. As built, **reuse is byte-identical to decoding again**, so the cache cannot change what a record renders. That is the whole safety argument, and it's asserted directly in both the tool and the tests.

Keyed on `CosStream` object identity: a worker opens its document once per session and `CosDocument` memoises loaded objects, so a page's streams are the same objects across records. The trace confirms the affinity holds — page 2's three records all on `worker=2`, page 3's two on `worker=0`, page 29's two on `worker=2`.

## Result

`tool/bench_record_reuse.dart` serializes each page twice with a shared cache and twice without, and checks the bytes:

```
   firstMs  secondMs  reuseMs   saved  identical  page
      57.9      17.0       4.5     74%        yes  p1
     251.8     224.6      10.1     95%        yes  p2
     222.4     214.2      15.4     93%        yes  p3
      17.7      25.4       4.4     83%        yes  p8
     244.3     230.6       9.8     96%        yes  p29
       8.1       7.4       2.8     62%        yes  p30
       8.9      10.7       2.9     73%        yes  p31

  re-record uncached 729.8ms -> cached 49.9ms (93% less),
  first record 811.2ms either way.
```

The saving lands exactly on the pages the trace flagged — 2, 3, 29 — at 93–96%. A page's first record is unchanged; this only removes repetition. Byte-identical on every page.

## Testing

Eight unit tests on the cache (reuse, size-keyed separation, native as its own key, identity not equality, declines uncached, LRU eviction, oversize bypass, clear), plus one codec-level test pinning the integration — the unit tests would all pass against a codec that never consulted the cache it was handed. That test was confirmed to fail against exactly that mutation.

pdf_graphics 966, dart_pdf_editor 1989, analyzer clean.

### DIFF
--- a/.github/workflows/deploy-app-web.yml
+++ b/.github/workflows/deploy-app-web.yml
@@ -50,7 +50,9 @@ jobs:
           --out ../packages/dart_pdf_editor_assets/assets/web/pdf_render_worker.dart.js
 
       # CanvasKit renderer (matches the app preview + release web builds).
-      - run: flutter build web --release
+      # PDF_BUILD_COMMIT stamps the build into the `?perf=1` trace, so a device
+      # trace says which build produced it instead of being dated by hand (#451).
+      - run: flutter build web --release --dart-define=PDF_BUILD_COMMIT=${GITHUB_SHA::7}
         working-directory: app
 
       - name: Cache-bust web build
--- a/.github/workflows/preview-app-web.yml
+++ b/.github/workflows/preview-app-web.yml
@@ -46,8 +46,15 @@ jobs:
           dart run dart_pdf_editor:build_web_worker
           --out ../packages/dart_pdf_editor_assets/assets/web/pdf_render_worker.dart.js
 
-      - run: flutter build web --release
+      # PDF_BUILD_COMMIT stamps the build into the `?perf=1` trace, so a device
+      # trace from a preview says which build produced it (#451). A preview
+      # builds the PR head, so the merge SHA would be the wrong answer.
+      - run: >-
+          flutter build web --release
+          --dart-define=PDF_BUILD_COMMIT=${PREVIEW_SHA::7}
         working-directory: app
+        env:
+          PREVIEW_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
 
       - name: Cache-bust web build
         run: bash tool/web_cache_bust.sh app/build/web
--- a/app/lib/main.dart
+++ b/app/lib/main.dart
@@ -14,6 +14,13 @@ import 'l10n/app_localizations.dart';
 import 'app.dart' deferred as app;
 import 'app_info.dart' deferred as app_info;
 
+/// The git commit this build was compiled from, or `unknown` for a build that
+/// did not pass one. Supplied by the build scripts as
+/// `--dart-define=PDF_BUILD_COMMIT=$(git rev-parse --short HEAD)`; a plain
+/// `flutter run` leaves it unset, which is honest rather than misleading.
+const kBuildCommit =
+    String.fromEnvironment('PDF_BUILD_COMMIT', defaultValue: 'unknown');
+
 /// On Windows and Linux the OS launches the app with the opened file as a
 /// command-line argument; the Flutter runner forwards it here.
 Future<void> main(List<String> args) async {
@@ -28,6 +35,12 @@ Future<void> main(List<String> args) async {
   // native, where there's no query string), so no `package:web` import.
   if (Uri.base.queryParameters['perf'] == '1') {
     PdfPerfLog.enabled = true;
+    // Stamp the build the trace came from, first line. A device trace is
+    // otherwise unattributable: a #451 trace captured 4 minutes after a fix
+    // commit turned out to predate the build and read as "the fix did
+    // nothing", which cost a round trip to work out (see
+    // doc/dev-log/2026-07-26-record-image-decode-reuse-451.md).
+    PdfPerfLog.log('build commit=$kBuildCommit');
   }
 
   runApp(_DeferredApp(launchArgs: args));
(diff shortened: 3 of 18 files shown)


ben-milanko:
## The first version was inert — device trace, and the fix

The exact-match cache did nothing on device. Page 2 still paid ~880 ms twice:

```
[29994] p2 worker=2 serialize=876ms  decodedMpx=0.4  cmyk:1  transcript=miss
[32677] p2 worker=2 serialize=884ms  decodedMpx=1.4  cmyk:1  transcript=hit
```

Same page, same worker, same declined CMYK image. Note `decodedMpx` — 0.4 against 1.4. A page's repeat records are a full-page pass and a 128 px thumbnail tile, **at different image pixel ratios by construction**, so they never ask for the same target and exact match never fires on the records that matter.

The fault was in my bench: it re-recorded at the *same* ratio, a sequence production never performs. `bench_record_reuse.dart` now records at a page ratio then a tile ratio.

### 1. Reuse across ratios, safely

A retained **native** decode may be downsampled to serve a smaller target — but only for streams whose decoder ignores the target anyway (`pdfImageDecodeIgnoresTarget`, i.e. DCTDecode). That isn't a concession: `bench_image_scale_cost` measured `tile/full = 1.01` for those, because `decodePdfImage(target)` on a DCT stream already *is* `downsamplePdfDecodedPixels(decodePdfImagePixels(...), tw, th)`.

The **native-entry** restriction is load-bearing — downsampling an entry that was itself a downsample would not match. Flate/CCITT/JPX keep the strict exact-match rule, because serving them from a downsample *would* change pixels.

### 2. The browser-codec decode was entirely uncached

`_withBrowserDecodedImages` ran `_decodeWithBrowserCodec` fresh on every record, at native resolution with no target, so it repeated identically:

```
p6   decode=318ms, 329ms, 331ms      p13  decode=537ms, 548ms, 631ms
p1   decode=421ms, 427ms             p5   decode=426ms, 424ms
p11  decode=421ms, 431ms             p23  decode=435ms, 406ms
p25  decode=437ms, 430ms             p4   decode=459ms, 439ms
```

Roughly 3–4 s of pure repeat in one scroll window — larger than the CMYK serialize total this PR was opened for. No target means no size dimension, so it keys on stream alone. Reuse reports as `reused=N` beside `codec=N`, so the next trace shows it directly.

## Result, on the pattern that actually occurs

```
    pageMs    tileMs  tileCachedMs   saved  identical  page
     105.4      12.1       10.5      13%        yes  p1
     298.9     224.2        9.1      96%        yes  p2
     223.6     220.5       21.3      90%        yes  p3
      37.8      46.6       15.2      67%        yes  p8
     287.5     249.4       22.9      91%        yes  p29
      14.5      14.6        6.7      54%        yes  p30
       9.2      30.1       26.6      12%        yes  p31

  thumbnail record after a full-page record: 797.5ms -> 112.4ms (86% less)
```

Byte-identical on every page. Pages 1 and 31 barely move — their images are cheap and there's little to reuse.

## Testing note

Both codec tests needed a second attempt, both times because **the fixture was too small to trip the resolution cap**. A 2×2 image decodes identically at every ratio, so the cross-ratio test passed against a codec that applied the native route to *every* format. It now uses a 64×64 gradient where ratio 0.2 binds, and asserts the routing structurally (a non-DCT stream must never be retained natively) rather than hoping two formats disagree on pixels. Confirmed to fail against the dropped-predicate mutation.

pdf_graphics 970, dart_pdf_editor 1989, zero Ghent baseline movement, analyzer clean.

Issue #647, opened 2026-08-04
[Bug]: Selection rectangle for proportional-font text doesn't sit on the actual glyphs


SpanManX:
### Affected package

pdf_graphics

### What happened?

Double-clicking a word in a PDF rendered by dart-pdf gives a selection rectangle that's visibly misaligned with the glyphs. For a string like Hello, world! in 24pt Helvetica, double-clicking world lands a highlight band that is offset by roughly 1/3 of the run width — well clear of the glyphs themselves.

### Minimal reproduction

# Windows
<img width="729" height="468" alt="Image" src="https://github.com/user-attachments/assets/d9719052-8b2a-4f12-a67a-192e3b9e41bd" />

<img width="545" height="201" alt="Image" src="https://github.com/user-attachments/assets/46ac7d73-c784-4c4a-8feb-0fd55ac0f86d" />

# Android

<img width="273" height="72" alt="Image" src="https://github.com/user-attachments/assets/12656e1c-fe33-4703-b01b-1f53c3ffafcb" />

<img width="330" height="228" alt="Image" src="https://github.com/user-attachments/assets/4b382ea2-1b09-4de8-9aac-99a75980780b" />

### Error output or stack trace

```text

```

### Version or commit

main

### Platform

Flutter Windows

### Environment details

```text

```

### Checks already run

```shell

```

### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I can provide a non-confidential reproducer or enough detail to create one.

ben-milanko:
Resolved by #648 and #655. #648 fixed the range-to-selection geometry for proportional fonts, while #655 fixed the remaining substituted-glyph painting drift tracked in #649. Together they cover the reported selection misalignment on non-embedded Helvetica, so I’m closing this as completed.

PR #648, opened 2026-08-05
Put selection geometry on the glyphs of proportional fonts (#647)

ben-milanko:
Refs #647 — deliberately **not** `Fixes`, so merging this leaves the issue open.

#647 has two independent halves. This PR fixes the first: the selection's own geometry (range → rectangle), which is now exact. The second is glyph *painting* — for a substituted font the device matches only the run's **total** width, so interior glyphs drift up to ~4pt mid-run and a correctly placed band still reads as offset. That half is split out as #649, with the measurements. #647 should close when both have landed.

## Summary

`PdfPageText.quadsFor` sliced a run's width by character **count**, which assumes every character is the same width. In a proportional font they are not, so the selection band sat beside the glyphs rather than on them. It had carried a placeholder since the beginning:

```dart
// approximate within-run positions by character fraction; per-glyph
// geometry arrives with the font engine
```

`positionNear` made the same assumption in reverse, so a double-click's two halves — point → character, character → rectangle — compounded the error.

On the issue's own repro (`Hello, world!` in 24pt Helvetica) the `world` band landed 3pt right and 6.5pt narrow. The error scales with how uneven the run is: `I` is 0.278em in Helvetica and `W` is 0.944em, so on `IIIIII WWWWWW` the highlight misses by more than a quarter of the run — the "roughly 1/3" the issue reports.

**The fix** carries the real advances end to end. The interpreter already accumulates the exact pen position per glyph; it just discarded it for anything but embedded outlines.

- `PdfTextRun.charOffsets` — em-space offset of every character boundary, length `text.length + 1`, last entry `== width`.
- `PdfInterpreter(collectCharOffsets:)` — **off by default**. Painting walks get the same positions from `PdfTextRun.glyphs` and would otherwise pay a list per run for nothing; `PdfTextExtractor` turns it on.

The flag is what makes the reported case work at all: `glyphs` is only built for `hasOutlines || isType3`, so a **non-embedded** standard-14 Helvetica — exactly the issue's PDF — has no glyph list. `charOffsets` needs only the metrics, so it covers substituted fonts too.

`PdfExtractedRun.charOffsets` carries the table to `quadsFor`/`positionNear`, which now binary-searches for the nearest boundary instead of scaling a fraction.

API impact: two new optional fields and one new optional constructor parameter. No existing signature changes; runs without a table fall back to the previous interpolation.

## Affected packages

- [ ] `pdf_cos`
- [ ] `pdf_document`
- [x] `pdf_graphics`
- [x] `dart_pdf_editor` (test only)
- [ ] `pdf_test_fixtures`
- [ ] Example app

## Change type

- [x] Bug fix
- [ ] Feature
- [ ] Rendering change
- [ ] Editing behavior change
- [ ] Performance change
- [ ] Refactor / cleanup
- [ ] Tests / fixtures only
- [ ] Documentation only

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze` — clean
- [x] `cd packages/pdf_cos && fvm dart test` — 356 pass
- [x] `cd packages/pdf_document && fvm dart test` — 808 pass
- [x] `cd packages/pdf_graphics && fvm dart test` — 1009 pass
- [x] `cd packages/dart_pdf_editor && fvm flutter test` — 2114 pass
- [ ] Example app smoke test

The render path is untouched — the flag leaves `charOffsets` null for every painting walk — so no render baseline can move. Covered instead by the deterministic counter gate (`tool/perf.sh gate`: 12 inputs, 13 counters within 3%) and the corpus A/B below, which does exercise the real 16-file corpus through parse + interpret + extract.

## Performance

Text extraction is on the changed path, so it was measured rather than eyeballed. `tool/perf.sh diff d84479b SCENARIO`, 4 interleaved runs:

| metric | `type3-text-sweep` | `dartpdf-corpus` |
| --- | --- | --- |
| `openMs` | 1.031x | 0.976x |
| `firstPageMs` | 0.934x | 0.991x |
| `interpretMs` | 0.953x | 0.979x |
| `extractMs` | 1.037x | 1.009x |
| `saveMs` | 1.031x | 0.990x |
| `peakRssBytes` | 1.033x | 1.017x |

Both VERDICT OK.

**The corpus sweep caught a regression the single-scenario run missed.** `extractMs` first came back at **1.114x — REGRESSED**, over the 1.10 gate, worst on `text-report-40p.pdf` at 1.45x. Two causes, both fixed in the first commit:

- `_sliceCharOffsets` copied the whole advance table per run, when the common case is one extracted run per source run and the slice is the whole thing already rebased at 0. It now shares the source list — nothing mutates it — saving an allocation and a copy per run.
- the per-glyph clamp read `charOffsets.last` (a bounds-checked call once per glyph); the tail is now tracked in a local.

`type3-text-sweep` alone said 1.037x and looked fine, so for extraction changes the corpus scenario is the one that matters. Also worth stating plainly: the first 1.114x was measured while I had sources open mid-run, which `CLAUDE.md` forbids for exactly this reason, so I discarded it and re-ran clean. Both columns above are complete uninterrupted runs.

## Notes for reviewers

Gotchas worth a look:

- **The BiDi pass invalidates the table.** `_bidiLine` reverses runs and re-splits them per glyph, so the offsets stop lining up with a piece's characters. `_sliceCharOffsets` declines unless the piece is the source's own characters in their own order, and `quadsFor`/`positionNear` additionally ignore a table on an RTL run. RTL keeps the previous interpolation exactly.
- **Vertical writing mode gets null** — the pen advances along y, so a per-character *x* offset is meaningless.
- **A /ToUnicode ligature** maps one code to several characters. The PDF positions the code, not its pieces, so the advance splits evenly; the invariant that matters is the `text.length + 1` length, asserted at the emit site.
- **Offsets are clamped non-decreasing.** A negative `Tc` tighter than a glyph's own width walks the pen backwards, and consumers binary-search these. `_bidiLine` already did the same `math.max` for glyph offsets.
- **Two codecs, not one — and both their tests passed with the offsets dropped.** `serializePageText` (`render_command_codec.dart`) is the render-worker hop, the path this fix actually travels in production; its test compared only search *hit counts*. `pdfEncodePageText` (`text_cache.dart`) is the on-disk tier (magic `PTX2` → `PTX3`); its sample runs carried no `charOffsets` at all. Neither would have noticed a field missing from the wire. Both now compare geometry, and the worker one was verified to fail with the field removed.

Three existing tests pinned the old approximation by name ("interpolated rects", a `perChar` expectation) and were rewritten to the exact metrics; one now also asserts the old value is >2pt away, so the approximation cannot quietly return.

Two lines remain uncovered by choice: the vertical-writing bail-out, and the partial-slice rebase that only runs for per-glyph LTR pieces inside a bidirectional line. A mixed-script fixture for the latter looked disproportionate given RTL geometry is unchanged here, but say the word and I'll pin it.

Session notes: `doc/dev-log/2026-08-05-per-character-advances-647.md`.

### DIFF
--- a/doc/dev-log/2026-08-05-per-character-advances-647.md
+++ b/doc/dev-log/2026-08-05-per-character-advances-647.md
@@ -1,0 +1,75 @@
+# Per-character advances for selection geometry (#647)
+
+Selection and search highlights sat beside the glyphs instead of on them.
+`PdfPageText.quadsFor` had carried a placeholder since the beginning:
+
+```dart
+// approximate within-run positions by character fraction; per-glyph
+// geometry arrives with the font engine
+final logicalStart = (overlapStart - run.startIndex) / run.text.length;
+```
+
+Slicing a run's width by *character count* assumes every character is the
+same width. In a proportional font they are not. On the `buildClassicPdf`
+fixture - `Hello, world!` in 24pt Helvetica, the issue's own repro -
+double-clicking `world` produced a band 3pt to the right of the word and
+6.5pt too narrow. The error scales with how uneven the run is: `I` is
+0.278em and `W` is 0.944em, so on `IIIIII WWWWWW` the highlight misses by
+more than a quarter of the run, which is the "roughly 1/3" the reporter saw.
+
+Both halves of selection were wrong and compounded: `positionNear` mapped a
+click to a character by the same linear fraction, then `quadsFor` mapped it
+back.
+
+## The fix
+
+The interpreter already accumulates the exact pen position per glyph - it
+just threw it away for anything but embedded outlines. Now it can keep it:
+
+- `PdfTextRun.charOffsets` - em-space offset of every character boundary,
+  length `text.length + 1`, last entry `== width`.
+- `PdfInterpreter(collectCharOffsets:)` - **off by default**. Painting walks
+  get the same positions from `PdfTextRun.glyphs` when the font is embedded
+  and would otherwise pay a list per run for nothing;
+  `PdfTextExtractor._interpret` turns it on.
+
+The flag is what makes this work for the reported case at all. `glyphs` is
+only built for `font.hasOutlines || font.isType3`, so a **non-embedded**
+standard-14 Helvetica - exactly the issue's PDF - has no glyph list.
+`charOffsets` needs only the metrics, so it covers substituted fonts too.
+
+`PdfExtractedRun.charOffsets` carries the table through to
+`quadsFor`/`positionNear` (which now binary-searches for the nearest
+boundary instead of scaling a fraction).
+
+## Gotchas
+
+- **The BiDi pass invalidates the table.** `_bidiLine` reverses runs and
+  re-splits them per glyph, so the offsets no longer line up with the
+  piece's characters. `_sliceCharOffsets` returns null unless the piece is
+  the source's own characters in their own order (`sourceEnd - sourceStart
+  == text.length`, and not RTL). RTL keeps the old interpolation exactly -
+  it already had per-glyph geometry from its own path.
+- **Vertical writing mode gets null**: the pen advances along y there, so a
+  per-character *x* offset is meaningless.
+- **One code can map to several characters** through a /ToUnicode ligature.
+  The PDF positions the code, not its pieces, so the advance splits evenly
+  across them - the invariant that matters is that the table stays
+  `text.length + 1` long (asserted at the emit site).
+- **Offsets are clamped non-decreasing.** A negative `Tc` tighter than a
+  glyph's own width walks the pen backwards; consumers binary-search these.
+  `_bidiLine` already did the same `math.max` for glyph offsets.
+- **Two codecs, not one.** `serializePageText` in `render_command_codec.dart`
+  is the render-worker hop - the app extracts text off-thread, so *that* is
+  the path the fix actually travels in production; missing it would have
+  left the bug live everywhere but tests. `pdfEncodePageText` in
+  `text_cache.dart` is the on-disk tier (magic bumped `PTX2` → `PTX3` so
+  stale blobs miss rather than mis-parse).
+
+## Measurements
+
+`tool/perf.sh gate`: 12 inputs, 13 counters within 3%.
+`tool/perf.sh diff d84479b type3-text-sweep` (~7400 glyphs/page, the densest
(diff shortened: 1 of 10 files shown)


ben-milanko:
Closing out the two things the description left open.

**Corpus perf A/B — the `extractMs` regression is resolved.**

`tool/perf.sh diff d84479b dartpdf-corpus`, 4 interleaved runs over 16 files, on an otherwise idle machine:

| metric | before the fixups | now |
| --- | --- | --- |
| `openMs` | 1.016x | 0.976x |
| `firstPageMs` | 1.005x | 0.991x |
| `interpretMs` | 0.995x | 0.979x |
| `extractMs` | **1.114x — REGRESSED** | **1.009x** |
| `saveMs` | 0.991x | 0.990x |
| `peakRssBytes` | 1.013x | 1.017x |

VERDICT: OK. The 11% came from two things, both now fixed in the first commit: `_sliceCharOffsets` copied the whole advance table per run when the common case is one extracted run per source run (it now shares the already-rebased list, since nothing mutates it), and the per-glyph loop re-read `charOffsets.last` instead of tracking the tail in a local. What's left is inside noise.

Worth noting explicitly: the earlier 1.114x number was measured while I had edited sources mid-run, which `CLAUDE.md` forbids for exactly this reason, so I discarded it and re-ran clean rather than trust it. Both figures above come from complete uninterrupted runs.

**Full `dart_pdf_editor` suite: 2114 pass** against the final tree, including the late non-decreasing-clamp tweak the description flagged as still pending.

**Coverage.** The first codecov pass flagged `text_cache.dart` at 44% on the patch — the codec's sample runs carried no `charOffsets`, so the new encode/decode branches were never exercised. That is a persistence path, and a silent drop there would revert highlights to the interpolated positions *only on a reopened document*, which is precisely the kind of bug that survives a green suite. Covered in fb7669a; that file is now 100% and patch coverage is 97.94%.

Two lines remain uncovered, deliberately: the vertical-writing-mode bail-out, and the partial-slice rebase in `_sliceCharOffsets` that only runs for per-glyph LTR pieces inside a bidirectional line. Building a mixed-script fixture for the latter looked disproportionate given RTL geometry is explicitly unchanged by this PR — offsets are declined for RTL runs, so those paths keep the previous interpolation byte for byte. Happy to add one if you'd rather have it pinned.

---

PR #658, opened 2026-08-10
fix(viewer): keep visible neighbour pages out of the prefetch path (#657)

ben-milanko:
## Summary

Scrolling around a large-format scan or CAD sheet made the pages above and below the current one visibly **pop in and out** as they sat near the screen edge.

Focus is a single page index — `focusDistance = |index - controller.currentPage|`, where `currentPage` is the page under the viewport **centre**. On any document whose pages are taller than the viewport (every CAD sheet, every A3 scan) there is always a *second* page filling much of the screen at distance 1. Three prefetch economies keyed off that distance and so applied to a page the user could plainly see:

1. `_imageRatioTarget()` decoded its images at `prefetchImagePixelRatioFactor` (0.5×, #451). On a scan the image *is* the page, so half the screen rendered soft.
2. `PdfLiveRasterBudget.rebalance()` evicted farthest-distance-first and protected only distance 0, so the visible neighbour was the first casualty once the on-screen pair exceeded the 384 MB desktop ceiling — two large-format base rasters plus their retained scenes get there easily. `evictLiveRaster()` drops the base raster, detail patch and scene, leaving the soft preview: the pop-out.
3. `evictReclaimable()` shed *every* non-focused page, visible ones included, on each platform low-memory signal.

Crossing the midpoint flips which page gets which treatment, so it oscillates. Because `PdfImageCache` keys by content *and* decoded size (`PdfSizedImageKey`), the 0.5× and 1.0× decodes of the same scan are two entries — each flip is a guaranteed miss, a fresh multi-MB decode, and an eviction of whatever the other page held. A devtools export from the reported session shows exactly that: decoded-image cache holding 11 entries / 130 MB with **210 misses against 146 evictions**, render-record cache at 67/100 MB with 201 evictions.

### The fix

`PdfPageView.onScreen` — whether *any part* of the page overlaps the viewport — now gates the prefetch economies instead of `focusDistance`:

- full-resolution image decode for every on-screen page; the reduction still applies to genuinely off-screen cache-window neighbours, which is what #451 was about;
- `rebalance()` runs two passes over the same farthest-first order — every off-screen page first, and only if the total is *still* over budget does it reach the visible ones (zoomed far enough out the visible set alone can exceed the ceiling, so the valve stays open, it just isn't the first move);
- `evictReclaimable()` spares on-screen pages: they re-rasterize the instant they are dropped, so evicting them under pressure buys a flicker and a *higher* peak — the same feedback loop `adaptive_memory.dart`'s `registryMaterialityPercent` comment documents;
- `didUpdateWidget` upgrades a reduced buffer when a page scrolls into view, not when it becomes the centre page.

`_updateCurrentPage` computes the visible span in the pass it already made over the page offsets. The span is a `Listenable` (`_PdfOnScreenSpan`) that each `_PdfViewerPage` subscribes to individually, so a crossing rebuilds only the pages that crossed — see the perf note below for why that matters. `currentPage` keeps its exact old rule, so page numbering and render focus are untouched.

Full write-up in `doc/dev-log/2026-08-10-visible-page-prefetch-demotion.md`.

**API impact:** `PdfLiveRasterHolder` gains `liveRasterOnScreen`; implementations must add it. `PdfPageView.onScreen` defaults to `true`, so a directly-embedded page view is unaffected.

## Affected packages

- [ ] `pdf_cos`
- [ ] `pdf_document`
- [ ] `pdf_graphics`
- [x] `dart_pdf_editor`
- [ ] `pdf_test_fixtures`
- [ ] Example app

## Change type

- [x] Bug fix
- [ ] Feature
- [ ] Rendering change
- [ ] Editing behavior change
- [x] Performance change
- [ ] Refactor / cleanup
- [ ] Tests / fixtures only
- [ ] Documentation only

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze`
- [ ] Example app smoke test

If any relevant check was not run, explain why: the change is confined to `dart_pdf_editor`'s viewer widgets and the live-raster budget, none of which the other packages' suites or the render baselines exercise — no rendering output changes, only when a page's raster and image decode are kept versus reclaimed. The full 2171-test `dart_pdf_editor` suite passes.

### Perf

Measured with the real-Chrome harness: **`tool/perf.sh webdiff 3cafba1 scroll-scan --iterations 7`**. Numbers and analysis in [this comment](https://github.com/ben-milanko/dart-pdf/pull/658#issuecomment-5236025021). Headline:

- `buildP50` +1.8%, `buildP95` +2.0% — flat.
- `buildMax` +37.9% (54.2 → 74.7 ms) — **a real regression, and a deliberate one.** It is exactly one frame per run (over-budget counts are unchanged run for run), the frame where a newly-visible page's full-resolution scene lands. Total work is *lower* than before — the old scheme rendered that page twice, soft then full — the peak just moved earlier, into the scroll instead of after the settle. Not tuned away, because an intermediate ratio would reinstate a smaller version of the softening this PR removes.
- The first cut of this PR regressed `buildP95` +16.8% by calling `setState` on the whole viewer for every span change; `731ed7c` fixed that (→ +2.0%) by making the span a `Listenable`.

### New tests

- `test/page_on_screen_test.dart` — a page sharing the screen with the current one stays on-screen; a prefetched page just past the edge does not; the span is measured on a document that never scrolls; the flag lands one frame after the scroll rather than at the settle; and `onScreen` is pinned to `PdfViewerController.visiblePageRegion` across a sweep of scroll offsets so the flag and the thumbnail strip's viewport indicator cannot drift apart.
- `test/live_raster_budget_test.dart` — four new ordering cases: a farther off-screen page goes before a nearer visible one, every off-screen page goes before any visible one, the safety valve reaches visible pages once the off-screen ones are gone, and the pressure path spares on-screen pages.

## Compatibility checklist

## Notes for reviewers

- **`focusDistance` is stale mid-scroll, by design and as it always has been** — it only reaches the page views on a viewer rebuild. That is now load-bearing rather than incidental: `onScreen` is delivered promptly and decides the image ratio and reclaim protection, leaving focus to do what it is good for, ordering evictions among pages nobody is looking at. A test that asserted fresh `focusDistance` was removed for this reason.
- **Testing gotcha:** the prefetch band is *offstage*. Pages inside the list's cache extent but past the viewport edge are laid out and not painted, so `find.byType(PdfPageView)` misses them — pass `skipOffstage: false` or "off-screen but mounted" looks like "not mounted at all".
- **Harness bug found along the way (not fixed here):** `app/tool/perf/driver.mjs:65` defaults Chrome to the macOS app path with no fallback, so `webdiff` cannot run unattended on Linux without an explicit `PERF_CHROME`. Happy to add a `PATH` fallback here or separately.

### DIFF
--- a/doc/dev-log/2026-08-10-visible-page-prefetch-demotion.md
+++ b/doc/dev-log/2026-08-10-visible-page-prefetch-demotion.md
@@ -76,15 +76,20 @@ numbering and render focus are untouched.
 
 Two things worth knowing:
 
-- **It rebuilds when the span changes.** Previously the page views only
-  learned their new `focusDistance` at the 200 ms transform settle or the
-  500 ms scroll settle, so a page crossing the edge kept its stale
-  prefetch treatment for the whole gesture and then re-rendered in a batch
-  at the settle - part of why the pop was so visible. The span changes at
-  most a couple of times per page crossing, so this is far cheaper than a
-  per-frame rebuild, and it lands the flag when it matters.
+- **The page views learn it when it changes, not at the next settle.**
+  Previously they only picked up their new `focusDistance` at the 200 ms
+  transform settle or the 500 ms scroll settle, so a page crossing the edge
+  kept its stale prefetch treatment for the whole gesture and then
+  re-rendered in a batch at the settle - part of why the pop was so
+  visible.
+
+  This first landed as a `setState` on the viewer, which measured badly
+  enough to be worth its own section - see "Measuring it, and the rebuild
+  that cost" below. The span ended up a `Listenable` each page subscribes
+  to instead.
 - **Scroll listeners can fire during layout**, so a mid-frame span change
-  defers its `setState` to a post-frame callback - the same hazard
+  defers to a post-frame callback before notifying (a listening page's
+  `setState` would be illegal there) - the same hazard
   `PdfViewerController._notifySafely` guards.
 
 ## Testing notes
--- a/doc/dev-log/2026-08-10-visible-page-prefetch-demotion.md
+++ b/doc/dev-log/2026-08-10-visible-page-prefetch-demotion.md
@@ -122,3 +122,42 @@ window at rest, which the full-width fixture cannot produce (at 800px wide
 each page is 1035px tall and page 1 is not built until it approaches). A
 300px-wide viewer makes each page 388px tall: pages 0 and 1 share the
 screen and page 2 sits below the fold, built and off-screen.
+
+## Measuring it, and the rebuild that cost
+
+`tool/perf.sh webdiff <base> scroll-scan` (12p A3 scanned grayscale - the
+scenario built to stress image decode and the image cache) on the first
+version of this fix, 3 interleaved runs:
+
+```
+buildMax      48.74 -> 73.94   +51.7%   ✗
+buildP95       3.73 ->  4.36   +16.8%   ✗
+buildP50       2.12 ->  2.18    +3.0%   ·
+```
+
+`openPageCountMs` also tripped the 1.1x threshold (468 -> 538, +15%), but
+per-iteration it is 459/538/468 against 473/538/664 - overlapping ranges,
+and nothing in this change runs before the page-tree walk. Noise.
+
+The build numbers were not noise. p50 flat with p95 and max moving is the
+signature of *occasional* expensive frames, and the cause was
+`_setOnScreenRange` calling the viewer's own `setState`. The span changes a
+couple of times per page crossing - far more often than the settle the page
+views used to rebuild on - and the viewer's `build` is enormous (the list,
+every mounted page's whole subtree, gesture recognizers, scrollbar,
+chrome). Rebuilding all of it to hand a bool to the two or three pages
+whose flag actually flipped is most of a frame on a dense sheet.
+
+So the span is a `Listenable` (`_PdfOnScreenSpan`), not viewer state: each
+`_PdfViewerPage` subscribes, recomputes its own answer, and setStates only
+when that answer flips. Same house pattern as `viewportChanges`, which
+exists for the same reason (the controller stays quiet during scrolling so
+the thumbnail strip's indicator does not spam every listener).
+
+One consequence worth stating, because a test asserted it before and no
+longer can: `focusDistance` still only reaches the page views on a viewer
+rebuild, so it stays *stale mid-scroll* - as it always has. That is fine,
+and is rather the point. Stale focus now costs nothing that matters:
+`onScreen` (delivered promptly) decides the image ratio and shields both
+visible pages from reclaim, and focus is left doing what it is actually
+good for - ordering evictions among the pages nobody is looking at.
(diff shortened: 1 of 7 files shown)


ben-milanko:
## Perf A/B: `scroll-scan`, real Chrome, 7 interleaved runs vs `3cafba1`

`tool/perf.sh webdiff 3cafba1 scroll-scan --iterations 7` — the 12p A3 scanned-grayscale scroll scenario, chosen because it is the one built to stress image decode and the image cache, which is what this change moves.

```
metric            baseline   current      Δ       verdict
────────────────────────────────────────────────────────────
buildP50             2.19      2.23     +1.8%     ·
buildP95             3.77      3.85     +2.0%     ·
buildMax            54.18     74.73    +37.9%     ✗ REGRESSION
openPageCountMs       482       497     +3.0%     ·
agentMemoryBytes  1.627e8   1.626e8     -0.0%     ·
frames                354       352     -0.6%     ℹ
```

### One real regression, and it is deliberate

**`buildMax` +37.9% is real.** At 7 samples it separates: current `63.6/65.4/69.6/74.7/77.6/81.1/87.5` against base `47.1/48.5/49.4/54.2/58.1/66.6/73.7` — the baseline's *worst* run sits below the current median. (I initially read this as noise at 3 iterations; that was wrong, and is why the run was repeated at 7.)

It is **exactly one frame**. Over-budget counts are unchanged run for run — 1–2 frames >16 ms, exactly 1 >32 ms, 0–1 >50 ms, in both arms — so nothing new crosses a threshold. The single worst build frame got ~20 ms worse.

That frame is where a newly-visible page's full-resolution scene lands, and on this scenario a page *is* one big A3 scan — 4× the pixels the 0.5× prefetch buffer used to replay. The work didn't appear from nowhere: the old scheme rendered that page **twice** (soft on approach, full on becoming focused), so total work is now *lower* — the peak just moved earlier, into the scroll rather than after the settle that followed the focus flip. The soft neighbour was cheap per-frame because it was wrong.

Left as-is rather than tuned away: an intermediate ratio for visible-but-not-focused pages would reinstate a smaller version of the softening this PR exists to remove. If the spike ever needs flattening, the lever is the existing motion render hold — let a page that just crossed the edge paint its preview and take the full-resolution render at the settle — not the resolution it eventually renders at.

### A regression that was mine, and is fixed

The first cut of the fix measured **buildP95 +16.8% and buildMax +51.7%** with p50 flat — the signature of occasional expensive frames. Cause: `_setOnScreenRange` called `setState` on the viewer. The span changes a couple of times per page crossing, far more often than the settle the page views used to rebuild on, and the viewer's `build` covers the list, every mounted page's subtree, the gesture recognizers, the scrollbar and the chrome.

`731ed7c` makes the span a `Listenable` instead: each `_PdfViewerPage` subscribes and rebuilds only itself, only when its own answer flips (the same pattern as `viewportChanges`). That took p95 from +16.8% to +2.0% and p50 to +1.8%.

`openPageCountMs` tripped the 1.1× threshold on both 3-iteration runs (+15.1%, +10.3%); at 7 iterations it settles to +3.0%, confirming it as noise — nothing here runs before the page-tree walk.

### Harness note

`webdiff` cannot run unattended off macOS: `app/tool/perf/driver.mjs:65` defaults `CHROME` to `/Applications/Google Chrome.app/...` with no fallback, so every iteration fails `✗ Chrome not found` after building both bundles. These runs needed an explicit `PERF_CHROME=/opt/pw-browsers/chromium-1194/chrome-linux/chrome`. Falling back to `chromium`/`google-chrome` on `PATH` would make the scenarios runnable in CI and in Linux dev containers — happy to do it here or as a separate PR, say which.

Full write-up in `doc/dev-log/2026-08-10-visible-page-prefetch-demotion.md`.

---

PR #675, opened 2026-08-13
Render scanned MRC pages: reach image codecs from mask decode, refine JBIG2 text symbols

ben-milanko:
## Summary

A scanned page laid out as mixed raster content — a JPX colour layer stencilled by a high-resolution JBIG2 text mask — rendered as a flat dark grey sheet with no legible text. Two stacked failures, both fixed here.

**1. The mask decode paths couldn't reach the image codecs.** `pdfImageStencilMask` and `pdfImageSoftMask` pulled their samples with `cos.decodeStreamData`, the plain stream-filter chain. JBIG2 and JPX are image codecs, not filters, so that throws `UnsupportedFilterException` — which both helpers caught and turned into `null`, meaning "no mask, paint the base opaque". On an MRC page that paints the dark colour layer edge to edge over the whole sheet. `decodePdfImageBase` has always routed those two through `Jbig2Decoder`/`JpxDecoder`; the mask paths never did.

Both now go through one shared `_maskSampleData` helper that returns the samples *plus the shape and depth the codec actually produced* — the codec, not the dictionary, is authoritative, since JPX carries its own dimensions and always yields 8-bit samples whatever `/BitsPerComponent` claims. The stencil path rejects anything that didn't come back 1-bit (§8.9.6.3 has no wider form).

**2. The JBIG2 decoder rejected refined text symbols.** With the mask reachable, `Jbig2Decoder.decode` still returned null: `FormatException: refined text symbols`. The stream refines most of its instances (SBREFINE=1) — a scanner shares one symbol dictionary and re-codes each instance's own ink against it, which is how it stays lossless on real scanned glyphs. T.88 §6.4.11 is now implemented in `_readTextRegion`: read SBRAT off the flags, decode IARDW/IARDH/IARDX/IARDY per refined instance, and refine against the symbol.

Two details worth review:
- The reference is offset, not aligned: region pixel (x, y) refines reference pixel (x − GRREFERENCEDX, y − GRREFERENCEDY), where GRREFERENCEDX = floor(RDW/2) + RDX. `_decodeRefinement` and the two context builders grew `dx`/`dy` for it; a generic refinement *region* segment passes 0, preserving its previous behaviour. Dart's `>>` is the floor division the spec asks for at negative deltas — `~/` is not.
- One refinement context set is shared by every refined instance in the region, so it is allocated outside the instance loop.

`curS` advances by the refined width, not the dictionary symbol's.

No API changes outside `pdf_test_fixtures` (`Jbig2Placement` gained optional `refined`/`refinedDx`/`refinedDy`; `grayJpxCodestream`/`grayJpxExpectedSamples` are new).

## Affected packages

- [x] `pdf_cos`
- [ ] `pdf_document`
- [x] `pdf_graphics`
- [ ] `dart_pdf_editor`
- [x] `pdf_test_fixtures`
- [ ] Example app

## Change type

- [x] Bug fix
- [ ] Feature
- [x] Rendering change
- [ ] Editing behavior change
- [ ] Performance change
- [ ] Refactor / cleanup
- [ ] Tests / fixtures only
- [ ] Documentation only

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze` (clean across the workspace)
- [x] `cd packages/pdf_cos && fvm dart test` (362 passing)
- [x] `cd packages/pdf_graphics && fvm dart test` (1037 passing)
- [ ] `cd packages/dart_pdf_editor && fvm flutter test`
- [x] Ghent corpus test or baseline update (`ghent_render_test.dart` — unchanged, no baseline update needed)
- [x] Real PDF corpus parse/render smoke test (the reporting document, plus `pdfjs_render_test.dart`, 164 passing)
- [ ] Example app smoke test

Also ran `tool/perf.sh gate`: 12 inputs, 13 counters within 3% — unchanged, as expected, since no corpus file previously reached either mask path with these codecs. The two suites left unchecked are untouched by this change; the render-facing ones (Ghent, PDF.js) were run instead.

The reproducing document is a customer drawing and is not attached. Both fixes are covered by programmatic fixtures instead:

- `pdf_cos/test/jbig2_roundtrip_test.dart` round-trips a text region mixing refined and verbatim instances, with RDW/RDH of both signs and non-zero RDX/RDY. The fixture JBIG2 encoder learned to emit refined instances (`Jbig2Placement.refined` + `_encodeRefinement`); its context layout mirrors the decoder's bit for bit, which is the whole contract between them.
- `pdf_graphics/test/image_pixels_test.dart` covers a JBIG2 stencil `/Mask`, a JPX `/SMask`, and a colour JPX `/SMask` (which correctly reads as "no mask" rather than inventing an alpha plane).
- The JPX codestream fixture moved to `pdf_test_fixtures` so both packages can use a real one.

## Compatibility checklist

## Notes for reviewers

Silent-opaque is the failure mode worth remembering here: a mask helper returning `null` for "I can't read this" is indistinguishable from "there is no mask", and turns an unreadable mask into an image that covers everything below it. If another mask codec ever shows up, it needs the same routing.

Still unsupported in the JBIG2 decoder, and unchanged by this PR: custom Huffman tables, refinement/aggregate symbol dictionaries (SDREFAGG=1), and refinement in Huffman-coded text regions. The class doc comment is updated to say so accurately.

One pre-existing behaviour I did not touch: `decodePdfImage`'s targeted path asks `pdfImageStencilMask` for a mask sized to the *base*, which on an MRC page downsamples a 4964px-wide text stencil to the colour layer's 1241px. The full-decode path (what the renderer used here) composites at the mask's resolution and stays sharp.

Background and the spec pointers are written up in `doc/dev-log/2026-08-13-mrc-jbig2-stencil-mask.md`.

https://claude.ai/code/session_01Xi7vLuBDLAF5Vyy8EPg1hi

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xi7vLuBDLAF5Vyy8EPg1hi)_


ben-milanko (commented):
One correctness issue in the new refined-symbol allocation guard; the codec-routing changes and the rest of the refinement implementation look sound.

packages/pdf_cos/lib/src/filters/jbig2.dart
```diff
@@ -831,13 +860,42 @@ class Jbig2Decoder {
...
+          final rdw = decoder.decodeInt(iardw) ?? 0;
+          final rdh = decoder.decodeInt(iardh) ?? 0;
+          final rdx = decoder.decodeInt(iardx) ?? 0;
+          final rdy = decoder.decodeInt(iardy) ?? 0;
+          final rw = symbol.width + rdw;
+          final rh = symbol.height + rdh;
+          // A corrupt stream can decode an arbitrarily large delta, so the
+          // allocation has to be bounded: no encoder codes an instance bigger
+          // than the region it is drawn into.
+          if (rw <= 0 || rh <= 0 || rw > w || rh > h) {
```
ben-milanko: [P2] Allow refined symbols to extend beyond the text region

Please don’t use the text region’s `w`/`h` as the allocation bound here. T.88 text-region composition clips symbol bitmaps at the region edge, so a valid refined instance may be wider or taller than the region (for example, a large glyph placed partly outside it). Reference decoders allocate `symbol.width + RDW` / `symbol.height + RDH` and clip during composition. With this check, `Jbig2Decoder.decode` returns null for such a stream and the entire page image is skipped. Keep the positive-size validation, but use an independent defensive allocation limit rather than `rw > w || rh > h`.

ben-milanko: Good catch — you're right, and I confirmed it in the code: `_drawSymbol` composes through `_Bitmap.compose`, which skips out-of-range rows and columns, so an oversized or overhanging instance clips rather than being invalid. Bounding by the region turned a legal stream into a dropped page image.

Fixed in 5e4ea63: the positive-size validation stays, and the region bound is replaced with an independent defensive ceiling on the instance — `_maxRefinedSymbolExtent` (2^20 per dimension) and `_maxRefinedSymbolPixels` (2^26) — which is what actually needs guarding, since RDW/RDH are decoded before anything is allocated. The per-dimension cap also keeps `rw * rh` from overflowing on the way to the area check.

Added a round-trip test for the case you described: a refined instance larger than the region in both axes *and* placed to overhang its right edge now decodes and clips to the same page the reference composition produces. It fails against the previous guard.

---

PR #676, opened 2026-08-13
fix(render): keep image /SMask compositing correct on Impeller

ben-milanko:
## Summary

A page dense with JPEGs rendered as black rectangles on Android in 3.5.0, while
the same build was correct on Android web.

`CanvasPdfDevice.drawImage` composited a deferred image `/SMask` with a single
paint carrying **both** `BlendMode.dstIn` **and** a red→alpha
`ColorFilter.matrix`. Skia honours both. **Impeller drops a draw paint's blend
mode when that same paint carries a colour filter**, so the filtered mask
composited `srcOver` instead — painting opaque black wherever the mask was
white, and leaving the base intact wherever it should have been cut away. Both
now live on a `saveLayer` paint, the shape `beginSoftMaskComposite` already used
for group soft masks (which is why `/SMask` *groups* were never affected).

Only images whose base comes back from the **platform JPEG codec** keep their
mask as a companion GPU surface (`_gpuSoftMasks`, from #659), so the failing set
was exactly the non-CMYK `DCTDecode` images — CMYK/Flate/Indexed/CCITT/JPX
decode in pure Dart with the mask already multiplied in. On web the worker
decodes colour JPEGs itself (libjpeg-turbo WASM / browser codec) and ships
pre-masked pixels, which is the whole of "broken on Android, fine on Android
web". On the reported page the split was stark: the CMYK logos rendered
perfectly while every DeviceRGB JPEG went black.

Fixing the canvas device exposed a second half of the same bug.
`FlutterGpuTileRasterBackend` — the **default wherever the platform supports
it**, and Android's manifest enables flutter_gpu — uploads one texture per image
and never consulted `pdfGpuSoftMaskOf`, so it painted the base opaque; a cut-out
sticker's JPEG is solid black outside the subject, so the tile stayed a black
rectangle. It now declines such a scene to the Canvas path, the same
conservative fallback it already uses for gradients, tiled patterns, and
non-normal blends. Corpus acceptance is unchanged (PDF.js 90/89, Ghent 16/41),
so no page in either suite has this shape — which is also why the GPU/Canvas
parity suite never caught it.

`flutter test --enable-impeller` reproduces the reported page pixel-for-pixel on
Linux, which also names the gap that let this ship: **every render test in CI ran
on Skia**, while the shipped Android app runs Impeller.

Background: [doc/dev-log/2026-08-13-impeller-image-soft-mask-black.md](doc/dev-log/2026-08-13-impeller-image-soft-mask-black.md).

## Affected packages

- [ ] `pdf_cos`
- [ ] `pdf_document`
- [ ] `pdf_graphics`
- [x] `dart_pdf_editor`
- [ ] `pdf_test_fixtures`
- [ ] Example app

(also `dart_pdf_editor_flutter_gpu`, which the list predates)

## Change type

- [x] Bug fix
- [ ] Feature
- [x] Rendering change
- [ ] Editing behavior change
- [ ] Performance change
- [ ] Refactor / cleanup
- [ ] Tests / fixtures only
- [ ] Documentation only

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze` (`--fatal-infos`, clean; lockstep constraint check passes)
- [x] `cd packages/dart_pdf_editor && fvm flutter test` (2298 passed)
- [x] Ghent corpus test or baseline update (44 passed, no baseline change)
- [x] Real PDF corpus parse/render smoke test (PDF.js suite, 164 passed)
- [ ] Example app smoke test

Also run: `app` suite (422 passed); `packages/dart_pdf_editor_flutter_gpu`
with and without `--enable-impeller --enable-flutter-gpu` (232 passed with a
live GPU context); the new compositing test under both engines. `pdf_cos`,
`pdf_document` and `pdf_graphics` are untouched by this change.

The reporter's document is not public, so both new tests build their fixture
programmatically: a solid-colour JPEG (`img.encodeJpg`) under a half-transparent
Flate grayscale `/SMask`, which is the exact shape that keeps a companion mask.

`test/image_soft_mask_composite_test.dart` asserts the masked-out half keeps the
page background. It passes on Skia with *either* paint shape, so CI now runs
that one file a second time with `--enable-impeller`; without the fix it fails
there with `Expected: [255, 255, 255, 255] Actual: [40, 200, 90, 255]` — the
inverted mask.

## Compatibility checklist

## Notes for reviewers

- `pdfGpuSoftMaskOf` is newly exported from `dart_pdf_editor` so the flutter_gpu
  backend can see the companion surface it must decline.
- The new CI job runs the flutter_gpu suite with the GPU flags. Every assertion
  in that package self-skips when the runner cannot open a GPU context, so it
  degrades to a no-op rather than a failure on a host without one.
- **Follow-ups, deliberately not in this PR:**
  - Broadening the Impeller CI job to more of the render suite. The engines
    differ by ~2/255 mean on antialiased content, so the pixel-baseline suites
    (Ghent, PDF.js) need their own tolerance before they can join it.
  - Recovering the GPU path for these pages by uploading the mask as its own
    texture (or multiplying it into the base's alpha in the readback branch of
    `_uploadImageTexture`, which already pays a `toByteData`).
  - The stencil (`/ImageMask`) branch of `drawImage` still sets
    `ColorFilter.mode(..., srcIn)` alongside `_elementBlend`, so Impeller will
    drop that blend when it is not `srcOver`. Left alone here: the layer rewrite
    that fixes the `/SMask` case would change knockout semantics (§11.4.5
    elements must replace only their own coverage, with no full-bounds layer),
    and no corpus page exercises the combination.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UA7oufbP7KSt7qRsL2Mb4q)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,6 +101,25 @@ jobs:
           token: ${{ secrets.CODECOV_TOKEN }}
           fail_ci_if_error: false
 
+      # Every other render test here runs on Skia, but the shipped Android app
+      # runs on Impeller - which honours a slightly different subset of the
+      # canvas API (issue #675: a draw paint's blend mode is dropped when the
+      # same paint carries a colour filter). Run the compositing tests that
+      # encode those invariants on Impeller too, so the divergence is caught
+      # here rather than on a device.
+      - name: Test dart_pdf_editor compositing on Impeller
+        working-directory: packages/dart_pdf_editor
+        run: flutter test --enable-impeller test/image_soft_mask_composite_test.dart
+
+      # The optional flutter_gpu tile backend is the default wherever the
+      # platform supports it (Android ships with it enabled), and its corpus
+      # suite asserts GPU/Canvas parity on every page it accepts. Every
+      # assertion self-skips when the runner cannot open a GPU context, so this
+      # is a no-op rather than a failure on a host without one.
+      - name: Test dart_pdf_editor_flutter_gpu
+        working-directory: packages/dart_pdf_editor_flutter_gpu
+        run: flutter test --enable-impeller --enable-flutter-gpu
+
       - name: Test pdf_ocr_vlm
         working-directory: packages/pdf_ocr_vlm
         run: flutter test --coverage
--- a/doc/dev-log/2026-08-13-impeller-image-soft-mask-black.md
+++ b/doc/dev-log/2026-08-13-impeller-image-soft-mask-black.md
@@ -1,0 +1,99 @@
+# Image /SMask composite turned black on Impeller (#675)
+
+Reported against 3.5.0 on Android: one page of a 62-page print PDF (a
+card-sheet page dense with JPEGs) rendered as black rectangles where the
+artwork should be. The same build on Android **web** was correct.
+
+## What was actually broken
+
+Not the decoder, not Android. `CanvasPdfDevice.drawImage` composited a
+deferred image /SMask with a single paint carrying **both**
+`BlendMode.dstIn` **and** a red→alpha `ColorFilter.matrix`:
+
+```dart
+canvas.drawImageRect(softMask, src, unit, Paint()
+  ..blendMode = BlendMode.dstIn
+  ..colorFilter = const ColorFilter.matrix(_redToAlpha));
+```
+
+Skia honours both. **Impeller drops the blend mode when the same draw paint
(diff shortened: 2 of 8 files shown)


ben-milanko (commented):
Reviewed the nested `saveLayer` composition, companion-mask lifetime through image clones, flutter_gpu fallback detection, regression fixtures, and the new Impeller CI coverage. I found no actionable issues. All three workflows are green at this head.

Issue #699, opened 2026-08-19
[Bug]: thumbnails and preview LODs re-interpret pages the viewer already holds (~2 s of platform thread per 10 s of scrolling)


ben-milanko:
## Affected package

`dart_pdf_editor`

## What happened?

Field trace of a 113-page corporate report being scrolled in the web app,
captured after #698's motion-safe lane landed. The lane itself is working -
17 grants read `cost=quiet`, `scene-cache reject` is gone, and the page the
reader lands on is ready ~9-70 ms after the hold lifts. What is left is that
**the two background surfaces cost more platform thread than the pages
themselves**, and they spend it re-doing work the viewer has already done.

Over the ~10 s window the trace covers:

| work | count | total | max |
|---|---|---|---|
| thumbnail tiles (`local interpret+raster`) | 27 | **1525 ms** | 92 ms |
| preview LOD warms (`prerender … full warm=`) | 9 | **523 ms** | 109 ms |
| page interprets (the actual document) | 12 | 584 ms | 77 ms |

and **49 of 58 janked frames are raster-dominated** (`JANK build=1.6ms
raster=88.0ms`), i.e. the dropped frames are these two rasterizing, not pages
building.

### 1. A 128 px thumbnail costs a full page interpret

With no active worker, `rasterizeThumbnail` falls through to
`PdfPageRenderer.renderImage(page, …)`, walking the whole content stream to
fill a 128 px tile - for a page the viewer interpreted seconds earlier and
still holds a retained scene for:

```
[perf 1005] scene-cache store page=74 commands=739 …
[perf 1337] thumbnail page=74 tile px=128 local interpret+raster=55.4ms (no worker)
```

### 2. The preview LOD ladder interprets the same page once per rung

`PdfPagePreviewCache._prerender` calls `renderImage` per LOD, so a page is
walked three times to produce base / 400px / 800px of the same content:

```
prerender page=107 lod=base  full warm=108.9ms
prerender page=107 lod=400px full warm=56.6ms
prerender page=107 lod=800px full warm=61.1ms
```

227 ms for one page, 174 ms for the next.

### Proposed fix

Build the picture **once** - preferring the page's retained scene when
`PdfPagePreviewCache` already has one, else one local walk - and rasterize
every size from it, instead of interpreting per tile and per LOD. That covers
both surfaces, needs no render worker, and is the largest remaining
main-thread cost on this journey.

### Two related findings from the same trace (split out if preferred)

[...]

PR #700, opened 2026-08-19
Build the page once for thumbnails and preview LoDs (#699)

ben-milanko:
## Summary

Fixes the two background surfaces in #699 that spent more platform thread than the document itself, by re-deriving work the viewer already had.

**The preview LoD ladder walked the page once per rung.** The rungs differ only in raster size, but `renderPreview` produced exactly one raster per call, so the viewer's prerender loop — one page, one rung at a time — interpreted page 107 three times to produce base/400px/800px (227 ms in the trace). `renderPreview` now takes `alsoFillLongestSides`: it collects the rungs that are missing, records the page **once** at the sharpest one's image resolution, and rasterizes every rung from that single `ui.Picture`. The viewer offers the still-missing rungs of a page inside the LoD window (`_ladderRungsFor`); vector-first motion passes stay base-only, since intermediates are image-complete by construction. A rung that rode along prints a `shared` token in its `prerender` line, and a `warm=` that is its own raster only.

Selection policy is unchanged — the nearest page missing a base preview is still picked first — but a base render for a page in the LoD window now also yields its sharper rungs, so a far page's base preview lands a couple of downscale rasters later than before. One existing test raced that ordering and now waits for both conditions; its assertions are intact.

**A 128 px thumbnail cost a whole page interpret.** With no active worker `rasterizeThumbnail` fell through to `renderImage`, walking the whole content stream to fill a tile for a page the viewer had interpreted seconds earlier and still held a retained scene for. It now takes the viewer's `PdfPagePreviewCache` and tries that scene first — ahead of the worker, which is a strictly more expensive way to obtain the same commands. The lookup has to know that "the same display" is not "an identical `PdfPageRenderPlan`":

* **Rotation** has two spellings (the page's own `/Rotate`, or null meaning it) and different producers write different ones.
* **Annotations** decide whether this fires in the app at all. The viewer bakes annotations into the page picture only when it is *not* drawing them in a live overlay layer — and an editing session, which is when the thumbnail strip exists, always uses the overlay. Its retained scenes are therefore annotation-free while the strip asks for annotations; on a page carrying none those are the same pixels, so only such a page tries the other spelling. Without this the hit rate in `PdfEditorView` would be zero.

Anything else — a view rotation, another paper colour, annotations on a page that has some, or images decoded below the tile's ratio — misses, and every path below is unchanged. The motion gate still applies: a replay is cheap but is still platform-thread work.

Also fixes the web harness's `prerender warms` counter, which has been reporting 0/0 since the ladder added its `lod=` token to that line.

## Affected packages

- [ ] `pdf_cos`
- [ ] `pdf_document`
- [ ] `pdf_graphics`
- [x] `dart_pdf_editor`
- [ ] `pdf_test_fixtures`
- [ ] Example app
- [x] Documentation / CI / repository metadata only (dev-log + `app/tool/perf/driver.mjs`)

## Change type

- [x] Bug fix
- [ ] Feature
- [ ] Rendering change
- [ ] Editing behavior change
- [x] Performance change
- [ ] Refactor / cleanup
- [ ] Tests / fixtures only
- [ ] Documentation only

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze` (clean, whole workspace)
- [x] `cd packages/dart_pdf_editor && fvm flutter test` (2358 pass)
- [ ] Example app smoke test

If any relevant check was not run, explain why: the change is confined to `dart_pdf_editor`'s preview/thumbnail caches and produces the same pixels through a different number of records, so the parser/render corpora are unaffected.

### Measurements

New `packages/dart_pdf_editor/test/benchmark_background_surfaces_test.dart` prices both halves on the VM and fails if either reuse stops paying:

```
background-surfaces ordinary:          ladder 19.1 -> 2.5 ms/page (-87%)    tile 2.2 -> 0.8 ms (-62%)
background-surfaces cad-vector:        ladder 920.1 -> 313.6 ms/page (-66%) tile 179.0 -> 92.5 ms (-48%)
background-surfaces letterhead-report: ladder 35.1 -> 16.6 ms/page (-53%)   tile 7.4 -> 5.5 ms (-25%)
```

On the VM the 128 px readback is most of a small page's tile, which is why the letterhead report gains 25% where the command-heavy CAD sheet gains 48%. On the web, where the trace measured a whole tile at 55 ms, the interpret is the bulk of it.

The two halves are gated differently on purpose: the ladder on wall time (it saves an interpret per rung, large on every profile), the tile on the interpret itself (`contentOps == 0`) — its ms column is real but close enough on a fast host that gating there would buy a flake rather than a signal. `page_preview_test` and `editing_thumbnail_cache_test` pin the mechanism the same way: the ladder tokenizes the page once where three per-rung calls tokenize it three times, and a tile served from a retained scene tokenizes nothing at all.

**Web A/B**, `tool/perf.sh webdiff e4dafb9 scroll-text --iterations 3` (`scroll-text` is the scenario that actually warms previews — `wheel` and `read` keep moving, and the prerender loop needs a full second of idle on web):

```
metric                  baseline    current     Δ         verdict
agentMemoryBytes        112953482   111146987   -1.6%     ·
buildMax                43.89       45.55       +3.8%     ·
buildP50                2.54        2.73        +7.5%     ·
buildP95                6.88        6.50        -5.5%     ·
jankCount               174         167         -4.0%     ·
openBytesMs             234         243         +3.8%     ·
openDocMs               238         247         +3.7%     ·
openPageCountMs         491         526         +7.1%     ·
✓ no metric regressed past 1.1×
```

The scenario has no metric for what a rung costs, so read the warm counts: full ladder rungs completed inside the same ~31 s window went **9/9/9 (baseline) → 11/14/13 (current)** — the same interpret budget producing about 40% more rungs. Caveats: this scenario is worker-backed, so the record was already off-thread and what a rung still costs there is its own CanvasKit readback; and the `open*` triple's ~4% drift is the harness, not this change (`openBytesMs` is fetching the file's bytes, which nothing here touches, and it moved with the other two). The thumbnail half has no web coverage — the perf harness has no thumbnail strip.

An earlier revision of this description quoted weaker tile numbers; that benchmark recorded its scene from a second `PdfDocument` over the same bytes, which misses on page identity, so both tile columns were timing the local interpret. Corrected in f125cda — see [this comment](https://github.com/ben-milanko/dart-pdf/pull/700#issuecomment-5341831769).

## Compatibility checklist

## Notes for reviewers

The issue's two other findings are deliberately untouched here, since both are separate from "build the page once":

* **The exact page-raster cache sitting inert** (163 `page-raster miss … reason=empty`, zero stores). The reporter's hypothesis holds up on reading: `putFullImage` is gated on `_renderedAtFullImageRatio()`, which refuses a page that rendered as a prefetch neighbour at half image resolution and then never re-rendered because its raster was already current.
* **The session having no render worker at all.** Worth confirming from the deploy's `webworker …` open lines, as the issue suggests.

Neither is worker-specific to this fix: the ladder's repeated walks cost the same on every platform, and the thumbnail reuse removes the interpret whether or not a worker exists.

Design notes and the reasoning behind the plan-matching rules are in `doc/dev-log/2026-08-19-thumbnail-preview-lod-reuse.md`.

### DIFF
--- a/app/tool/perf/driver.mjs
+++ b/app/tool/perf/driver.mjs
@@ -193,7 +193,12 @@ function parse(lines, frames) {
     }
     const warm = line.match(/worker warm=([\d.]+)ms/);
     if (warm) r.workerWarmMax = Math.max(r.workerWarmMax, Number(warm[1]));
-    const pre = line.match(/prerender page=\d+ (?:worker )?(vector|full) /);
+    // `prerender page=N lod=<rung> [worker ][shared ]<vector|full> warm=…`.
+    // The lod token arrived with the preview ladder and `shared` with #699's
+    // one-record-per-ladder build; both are optional so an older bundle's
+    // lines still count.
+    const pre = line.match(
+      /prerender page=\d+ (?:lod=\S+ )?(?:worker )?(?:shared )?(vector|full) /);
     if (pre) r.prerender[pre[1]]++;
     // The #527 bounded early prefix is real ink on the target page too, so it
     // counts as first content for the baseline (progressive off) - otherwise the
--- a/doc/dev-log/2026-08-19-thumbnail-preview-lod-reuse.md
+++ b/doc/dev-log/2026-08-19-thumbnail-preview-lod-reuse.md
@@ -1,0 +1,145 @@
+# Thumbnails and preview LoDs reuse work the viewer already did (#699)
+
+The field trace on #699 (113-page corporate report, scrolled in the web app
+after #698's motion-safe lane landed) put the two background surfaces above
+the document itself in platform-thread cost over a ~10 s window: 1525 ms of
+thumbnail tiles and 523 ms of preview-LoD warms against 584 ms of actual page
+interprets, with 49 of 58 janked frames raster-dominated. Both numbers were
+spent re-deriving something the viewer already had.
+
+## The preview ladder walked the page once per rung
+
+`PdfPagePreviewCache.renderPreview` produced exactly one raster per call, so
+the viewer's prerender loop - which selects one page and one rung at a time -
+walked the same content stream for base, then again for 400px, then again for
+800px:
+
+```
+prerender page=107 lod=base  full warm=108.9ms
+prerender page=107 lod=400px full warm=56.6ms
+prerender page=107 lod=800px full warm=61.1ms
+```
+
+The rungs differ only in raster size. `renderPreview` now takes
+`alsoFillLongestSides`: it collects the rungs that are missing, records the
+page **once** at the sharpest rung's image resolution, and rasterizes every
+rung from that one `ui.Picture`. A rung that rode along on another rung's
+record prints a `shared` token in its `prerender` line and a `warm=` that is
+its own raster only, so a trace still says what each rung cost.
+
+The viewer decides what rides along (`_ladderRungsFor` in pdf_viewer.dart):
+the still-missing intermediate rungs of a page inside
+`PdfPagePreviewLodPolicy.intermediateWindow`. Vector-first motion passes stay
+base-only - intermediates are image-complete by construction.
+
+**Selection policy is unchanged, arrival order is not.** The loop still picks
+the nearest page missing a base preview first; what changed is that a base
+render for a page in the tight LoD window now also yields that page's sharper
+rungs, since they cost only their own raster once the page is interpreted. A
+far page's base preview therefore lands a couple of downscale rasters later
+than before. `idle prerender promotes only the nearby LoD working set` in
+page_preview_test.dart had to wait on both conditions instead of racing them;
+its assertions (page 4 base-covered, page 4 without intermediates) still hold.
+
+## A 128px thumbnail cost a whole page interpret
+
+With no active worker `rasterizeThumbnail` fell through to
+`PdfPageRenderer.renderImage`, walking the entire content stream to fill a
+128px tile - for a page the viewer had interpreted seconds earlier and still
+held a retained scene for:
+
+```
+[perf 1005] scene-cache store page=74 commands=739 …
+[perf 1337] thumbnail page=74 tile px=128 local interpret+raster=55.4ms (no worker)
+```
+
+`rasterizeThumbnail` now takes the viewer's `PdfPagePreviewCache` and tries
+its retained scene first - ahead of the worker, which is a strictly more
+expensive way to obtain the same commands. A hit is a replay plus the tile's
+own (tiny) raster: no content-stream parse, no image codec, and a
+`thumbnail … retained replay+raster=` line in place of the
+`local interpret+raster=` one.
+
+What the lookup has to get right is that "the same display" is not the same as
+"an identical `PdfPageRenderPlan`". Two of the plan's inputs have spellings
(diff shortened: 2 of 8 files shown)


ben-milanko:
Measurements, promised in the description — plus a correction to the tile numbers it quotes.

### Web A/B (real Chrome)

`tool/perf.sh webdiff e4dafb9 scroll-text --iterations 3`. `scroll-text` is the scenario that actually warms previews: `wheel` and `read` keep moving, and the prerender loop needs a full second of idle on web, so neither ever starts it.

```
metric                  baseline    current     Δ         verdict
agentMemoryBytes        112953482   111146987   -1.6%     ·
buildMax                43.89       45.55       +3.8%     ·
buildP50                2.54        2.73        +7.5%     ·
buildP95                6.88        6.50        -5.5%     ·
jankCount               174         167         -4.0%     ·
openBytesMs             234         243         +3.8%     ·
openDocMs               238         247         +3.7%     ·
openPageCountMs         491         526         +7.1%     ·
✓ no metric regressed past 1.1×
```

The scenario has no metric for what a preview rung costs, so the warm counts are the thing to read: full ladder rungs completed inside the same ~31 s window went **9/9/9 (baseline) → 11/14/13 (current)** — the same interpret budget producing about 40% more rungs — while janked frames fell 4% and agent memory 1.6%.

Two caveats I'd rather state than have you find:

* This scenario is worker-backed, so the record was already off the platform thread; what a rung still costs there is its own CanvasKit readback, which no reuse removes. The worker-less session the issue traced is the bigger win.
* The `open*` triple drifting ~4% is the harness, not this change: `openBytesMs` is fetching the file's bytes, which nothing here touches, and it moved with the other two.

The thumbnail half has no web coverage at all — the perf harness has no thumbnail strip.

### Correction to the VM tile numbers

The benchmark table in the description under-reported the thumbnail half. Its retained scene was recorded from a second `PdfDocument` opened over the same bytes, which looks identical but misses every lookup, because the cache's freshness test is page *identity* — so both tile columns were timing the local interpret. Fixed in f125cda (record from `controller.pageAt(i)`, the object the tile asks about). Real numbers:

```
background-surfaces ordinary:          ladder 19.1 -> 2.5 ms/page (-87%)    tile 2.2 -> 0.8 ms (-62%)
background-surfaces cad-vector:        ladder 920.1 -> 313.6 ms/page (-66%) tile 179.0 -> 92.5 ms (-48%)
background-surfaces letterhead-report: ladder 35.1 -> 16.6 ms/page (-53%)   tile 7.4 -> 5.5 ms (-25%)
```

The ladder figures are unchanged; only the tile column moved. On the VM the 128 px readback is most of a small page's tile, which is why the letterhead report gains 25% where the command-heavy CAD sheet gains 48%.

That same commit also changes what the tile half is gated on: `contentOps == 0` (the interpret it removes) rather than wall time, whose margin is real but close enough on a fast host to buy a flake rather than a signal. The ladder half keeps its wall-time gate.

### Also here

`app/tool/perf/driver.mjs`'s `prerender warms` line has been reporting 0/0 on every scenario since the ladder added its `lod=` token to the format its regex matched. Fixed, which is how the counts above exist.

---

PR #726, opened 2026-08-22
perf(web): expose worker image-cache request deltas

ben-milanko:
## Summary

- capture the decoded-image cache after command serialization, including pure-Dart regional decoder activity that the previous pre-serialize snapshot missed
- add per-request cacheDelta counters beside lifetime cache totals
- aggregate hit/miss request paths and net cache growth in Patrol schema 6, globally and per scenario
- document the expanded Patrol evidence

## Investigation result

The local production Chrome CAD journey passed at 400% zoom with 128 tiles landed and 3 detail-scene adoptions. Corrected detail traces reported 0 hits, 2-3 misses, 0 new entries, and 0 bytes. Those misses line up with the visible ImageMask/non-DCT requests: the browser prepass declines them and serializeCommands performs the intended direct regional decode. Cache growth instead came from fallback full-record requests storing new stream/target-size combinations for previously unseen windows. Returning to the first window reused retained tiles and did not create another worker decode.

This points to expected region/format behavior rather than a broken cache key. Hosted Patrol should confirm the same request-level pattern.

## Validation

- fvm dart analyze
- fvm dart tool/patrol_perf_summary_test.dart
- fvm dart run dart_pdf_editor:build_web_worker --out /private/tmp/pr726-pdf-render-worker.js
- fvm flutter test test/render_worker_test.dart test/tile_image_detail_test.dart test/render_trace_gate_test.dart
- headless Chrome Patrol performance target at 2600x2600: both journeys passed

## Next investigation

Once this measurement lands, inspect detail payload binary packing. Local and hosted traces continue to show serialization/binning as the larger actionable cost; image-mask detail decline/fallback remains the following correctness-safe optimization candidate.

### DIFF
--- a/packages/dart_pdf_editor/example/README.md
+++ b/packages/dart_pdf_editor/example/README.md
@@ -50,10 +50,11 @@ These wall-clock measurements are review evidence rather than a pass/fail gate;
 compare the JSON with a recent `main` run from the same hosted-runner lane. The
 summary keeps global aggregates for continuity and a scenario breakdown for
 elapsed time, jank, reconciliation, raster work, worker phases, decoded-image
-cache pressure, and worker outcomes, so a slow heavy-document journey is not
-hidden inside the other browser journeys. The performance target includes an
-image-heavy, ultra-wide CAD scenario that deep-zooms and pans through distant
-regions, verifying worker-backed image detail and reusable tile delivery.
+cache pressure and per-request hit/miss deltas, and worker outcomes, so a slow
+heavy-document journey is not hidden inside the other browser journeys. The
+performance target includes an image-heavy, ultra-wide CAD scenario that
+deep-zooms and pans through distant regions, verifying worker-backed image
+detail and reusable tile delivery.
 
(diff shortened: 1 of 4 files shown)


ben-milanko:
Hosted Patrol confirms the local interpretation.

- Schema 6 captured 59 image-cache samples globally: 33 hits / 90 misses, proving reuse is active outside the CAD cold-window sequence.
- CAD captured 9 image-bearing worker requests: 5 detail and 4 record.
- Every detail request was 0 hits / 2-3 misses / 0 new entries / 0 bytes. The misses line up exactly with the 2 ImageMask draws, plus one non-DCT draw in the two declined regions; these then take the direct regional serializer path.
- Only full record requests grew the cache: 31 entries / 58.1 MiB total, as new distant windows introduced previously unseen stream/target-size combinations.
- The final revisit completed as detail without another fallback record.

Result: no evidence of a cache-key or lifecycle bug. The actionable costs remain detail bin/packing and the two ImageMask/non-DCT detail declines that trigger fallback records.

PR #727, opened 2026-08-22
perf(raster): cull offscreen shapes before flattening

ben-milanko:
## Summary

- treat wholly offscreen small fills and strokes as handled after the shape cache has already computed their conservative padded device bounds
- keep partially visible and oversized shapes on the existing direct unquantized fallback
- expose a viewport-cull counter and cover the no-fallthrough/cache-pollution behavior

This targets the deep-zoom CAD shape from #726: tens of thousands of tiny strokes distributed across a very wide page, with only one viewport visible. Previously each offscreen cache probe transformed the path, classified it as a bypass, then transformed and flattened it again in the direct path even though it emitted no strips.

## Performance evidence

- focused 30k-command CAD-shaped bin probe: median 68.7 ms -> 25.8 ms (62% lower), with identical emitted strip count
- local 2600x2600 Chrome Patrol performance target: both journeys passed; the final successful CAD detail bin was 190 ms
- #726 hosted-main evidence saw successful CAD detail bins at 568-842 ms; runner variance makes this PR hosted Patrol run the authoritative comparison

## Validation

- fvm dart analyze
- full pdf_graphics suite: 1,040 tests passed
- shape-cache + strip-plan focused tests: 24 passed
- render_worker_strips_test.dart, strip_plan_device_test.dart, tile_image_detail_test.dart, cad_wide_region_replay_test.dart: 31 passed
- production worker compiled with dart2js
- headless Chrome Patrol perf_e2e_test.dart at 2600x2600: 2 journeys passed

### DIFF
--- a/packages/pdf_graphics/lib/src/raster/strip_generator.dart
+++ b/packages/pdf_graphics/lib/src/raster/strip_generator.dart
@@ -239,15 +239,14 @@ class StripGenerator {
   /// Cached like [fillPath] when a [shapeCache] is attached; stroke
   /// parameters (device width, caps, joins, dashes - quantized to 1/64 px)
   /// join the content key.
-  void strokePath(
-      PdfPath path, PdfMatrix transform, PdfStroke stroke, int rgba,
+  void strokePath(PdfPath path, PdfMatrix transform, PdfStroke stroke, int rgba,
       {double tolerance = 0.25}) {
     if (path.isEmpty || _rowCount == 0) return;
     final cache = shapeCache;
     if (cache != null &&
         path.segments.length <= ShapeStripCache.maxSegments &&
-        _shapeCached(cache, path, transform, PdfFillRule.nonzero, stroke,
-            rgba, tolerance)) {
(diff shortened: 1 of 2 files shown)


ben-milanko:
Hosted Patrol comparison against exact main baseline cecba7ae (#726):

- Successful CAD detail-bin median: 670 ms -> 256 ms (-61.8%)
- Successful detail-bin range: 568-842 ms -> 219-323 ms
- Detail outcomes: 3 ok / 2 declined on main -> 5 ok / 0 declined on this PR
- CAD worker deliveries: 6 -> 9; local fallbacks: 2 -> 0
- Global worker-total p95: 1238.3 -> 969.8 ms (-21.7%)
- CAD worker-total p95: 1447.7 -> 1232.6 ms (-14.9%)
- Raster p95 was effectively flat (+0.2%)

The single-run journey wall time/jank was noisier (+6.1% / +13.6%), alongside unrelated heavy-document and raster variance. The isolated worker phase and completion-path changes match the focused 30k-command probe (68.7 -> 25.8 ms, -62%), so the targeted improvement is demonstrated without relying on total-journey noise. Both hosted journeys passed.

PR #730, opened 2026-08-22
perf: cull worker detail transcripts by viewport

ben-milanko:
## Summary
- select the paint units intersecting a deep-zoom viewport before worker image decode, command serialization, and strip binning
- cache the region index with the worker transcript and reuse it across pans on web and native
- retain the full-transcript fallback for unsupported state, broad regions, and selections that would not shrink
- report `regionCommands=selected/total` in web perf traces

## Why
The merged #729 Patrol trace moved reconciliation out of the hot path (2 ms p95). Deep-zoom worker details still spent 289-374 ms binning and 84-214 ms serializing because every viewport replay walked all 30,000 CAD strokes.

## Validation
- `fvm dart analyze`
- web render worker compiles with dart2js
- focused worker/transcript tests
- region replay/grid/worker and detail suites
- all 2,426 `dart_pdf_editor` tests
- new isolate regression proves the selective detail raster is byte-identical to the full-transcript region raster

The Patrol web perf artifact should validate the command-selection ratio and the expected bin/serialize reduction against merged #729.

PR #731, opened 2026-08-22
perf: stop edge pages refilling tile cache

ben-milanko:
## Summary

- keep a narrow edge-sliver page presentation-only once it falls below the viewer foreground-quality threshold
- let it paint retained exact/fallback tiles while preventing new visible misses and pan-ahead work
- cover the policy at the tile store, page-layer wiring, and real two-page viewer boundary

## Trace evidence

The field trace from merged #730 confirms viewport-culling is working: visible detail completed in roughly 222-323ms and page 2 first painted detail at about 298ms. The remaining tail was tile-cache churn across the page 1/page 2 boundary:

- 191 tile batches total
- 140 prefetch batches / 424 prefetch tiles
- page 1 prefetch: 196 tiles
- page 2 prefetch: 228 tiles
- shared tile cache stayed near its 100MB cap, so the two page pyramids repeatedly evicted and refilled one another

At that boundary page 1 still overlapped the viewport, but occupied less than the 15% foreground threshold. Page 2 was therefore the sole quality page; the old page-local policy still allowed page 1 to schedule because the global quality count was one. This change keys scheduling on the page own quality-visible state as well.

The CI Patrol CAD fixture is single-page, so it cannot reproduce this specific cross-page contention. The deterministic regressions assert the scheduling boundary directly; a future two-page field trace should show no new page 1 prefetch after page 1 becomes an edge sliver.

## Validation

- fvm dart analyze
- focused boundary/tile tests: 61 passed
- adjacent render/page/CAD/worker tests: 121 passed
- full dart_pdf_editor suite: 2,429 passed, 37 expected skips

### DIFF
--- a/packages/dart_pdf_editor/lib/src/pdf_page_view.dart
+++ b/packages/dart_pdf_editor/lib/src/pdf_page_view.dart
@@ -5038,6 +5038,12 @@ class _PdfPageViewState extends State<PdfPageView>
                 (_awaitingExactDetailPaint && !_tilePanAheadActive)
             ? 0
             : null,
+        // A narrow page sliver can keep presenting tiles it already owns, but
+        // it must not refill misses or a pan ring. Once the foreground count
+        // drops back to one, treating every mounted edge page as that sole
+        // claimant lets adjacent pyramids consume the shared LRU in turn and
+        // creates an endless raster/eviction/repaint loop.
+        scheduleMissing: widget.onScreen && widget.qualityVisible,
         // A retained picture keeps vector/text edges transform-sharp. An
         // upscaled coarser raster tile would cover that better base with a
         // visibly blurry square while the exact tile is pending. Coarse tiles
--- a/packages/dart_pdf_editor/lib/src/tile_layer.dart
+++ b/packages/dart_pdf_editor/lib/src/tile_layer.dart
@@ -36,6 +36,7 @@ class PdfTileLayer extends StatelessWidget {
     this.maxNewTilesPerPaint,
     this.maxInFlightTiles,
     this.prefetchRingOverride,
+    this.scheduleMissing = true,
     this.allowCoarserFallback = true,
     this.fallbackOcclusionFraction,
     this.onExactViewReady,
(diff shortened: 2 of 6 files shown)


ben-milanko:
CI and performance review are complete.

- All PR checks passed: full test/analyze, worker compile, Linux, Android/iOS/macOS Patrol, GPU previews, web demo Patrol/perf, deploy previews, and Codecov.
- Local validation: analyzer clean; 61 focused boundary/tile tests; 121 adjacent viewer/render/CAD/worker tests; full dart_pdf_editor suite 2,429 passed with 37 expected skips.
- Browser Patrol passed. Its one-baseline timings were slower than the immediately preceding main run, but all material values sit inside the previous six successful main-run envelope:
  - jank p95: 425.7 ms (main range 353.5-441.3)
  - worker total p95: 1,199 ms (865-1,552)
  - decode p95: 74 ms (53-126)
  - raster p95: 480.6 ms (435.8-526.6)
  - CAD journey: 16.2 s (12.3-17.5)
  - heavy-document journey: 9.31 s (6.92-9.47)
- Structural worker outcomes are effectively unchanged. The current Patrol CAD scenario is single-page, so it does not exercise the cross-page edge-sliver contention fixed here.

Merge assessment: no blocker found. The deterministic regressions cover the store policy, PdfPageView wiring, and the real two-page 15% foreground boundary.

Next follow-ups after merge:
1. Add a two-page deep-zoom CAD Patrol journey and summarize visible/prefetch tile counts per page, so this exact cache-contention class is measured on every PR.
2. Investigate moving-region detail coalescing: the #730 field trace completed three page-2 detail records (323/249/222 ms) while moving at roughly 2,800 px/s, but only the last painted.
3. Re-run the original two-page field journey and verify page 1 emits no new prefetch work after becoming an edge sliver.

ben-milanko:
Field trace follow-up (new trace vs the earlier #730 trace; both are 500-event observational captures):

- non-current edge-page tile work: **96 batches / 215 tiles → 0 / 0**
- all tile work: **191 batches / 497 tiles → 108 / 195** (43% fewer batches, 61% fewer tiles)
- prefetch work: **140 batches / 424 tiles → 16 / 52** (about 88% lower)
- JANK records: **67 → 34**; total-frame p50 **59.4 → 51.8 ms**, max **119.3 → 89.1 ms**; p95 is effectively flat/noisy (**77.8 → 80.8 ms**)
- page 2 still produces useful detail while crossing the boundary (first detail paint in 239 ms), and there is no missing-page/render starvation signature.

This is strong field confirmation that #731 stops the edge-sliver tile eviction/refill loop. The trace is unstamped, so the numbers remain observational rather than a CI gate. Remaining edge-page detail jobs and the 450 ms larger page-2 detail pass are good follow-up targets, not blockers for this PR. Current CI is green and the PR remains mergeable.

PR #747, opened 2026-08-23
ci(perf): repeat Patrol performance scenarios

ben-milanko:
## Outcome

Patrol performance comments now headline the median of three dedicated scenario runs instead of presenting one noisy journey as p95. Web, Android, and iOS collect three complete samples; phase and tail diagnostics remain p95.

The report shows main / PR sample counts and marks any scenario with fewer than three samples provisional. Current CI also rejects an incomplete PR sample set, so a native logging race cannot silently weaken the comparison.

## Changes

- repeat the web performance target three times on main and preview workflows
- repeat the native performance target three times on Android and iOS
- use scenario elapsed p50 in visible headlines while retaining p95 detail
- distinguish repetition counts from scenario coverage
- require every current scenario to produce all three complete start/end pairs
- retry iOS attempts whose passed XCTest run omits the Flutter completion marker
- document the sampling and baseline-transition contract

## Validation

- full repository analysis: clean
- Patrol summarizer tests: pass, including aggregation, sparse warnings, coverage, and completeness checks
- web PR run: all five scenarios produced 3 samples; medians stayed close to main
- Android PR run: 3 stable native tile samples
- iOS first PR run: 3 stable native tile samples
- observed iOS race: the new guard rejects the real 2-sample artifact; the retry path is exercised by CI
- workflow YAML parses and the Android runner passes bash syntax plus a mocked 1+3 invocation check

<details>
<summary>Baseline transition</summary>

This PR initially compares three PR samples against the latest one-sample main artifact and therefore labels timing deltas provisional. Once merged, the next main Patrol run publishes the first three-sample baseline; subsequent PR reports become like-for-like automatically.

</details>

### DIFF
--- a/.github/workflows/patrol-e2e.yml
+++ b/.github/workflows/patrol-e2e.yml
@@ -14,6 +14,7 @@ env:
   FLUTTER_VERSION: 3.47.0
   PATROL_CLI_VERSION: 4.7.0
   PATROL_ANALYTICS_ENABLED: "false"
+  PATROL_PERF_REPETITIONS: 3
   EXAMPLE_DIR: packages/dart_pdf_editor/example
   PDF_PATROL_BUILD_COMMIT: ${{ github.event.pull_request.head.sha || github.sha }}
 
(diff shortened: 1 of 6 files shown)


ben-milanko:
## Ready to merge

All required performance scenarios produced three complete PR samples. The median scenario time is lower than the current `main` baseline on web, Android, and iOS; this PR changes measurement repeatability rather than runtime behavior.

<details>

- Web: 3 complete runs for all five scenarios; elapsed medians range from -15.3% to +2.2% versus the current single-run baseline.
- Android: 3/3 complete `native-mobile-tiles` runs; median 3096 ms, -16.4% versus `main`.
- iOS Simulator: 3/3 complete `native-mobile-tiles` runs; median 2107 ms, -10.9% versus `main`.
- The completeness guard passed for web, Android, and iOS, so a Patrol/XCTest early-return race can no longer silently publish a partial comparison.
- Generated comments put headline comparisons first and keep the full trace tables collapsed.
- Static analysis, summary tests, workflow syntax checks, platform builds, and Patrol jobs are green.

The provisional warning is expected until `main` produces its first three-sample baseline after merge.

</details>

Issue #682, opened 2026-08-14
Make the Flutter GPU backend reliable on Flutter 3.47 and prove its value


ben-milanko:
## Summary

The native app already **requests `flutter_gpu` by default** when the platform exposes it, with automatic per-scene Canvas fallback. However, the custom backend currently accelerates only retained deep-zoom LoD tiles; normal page rasters, fast-scroll previews, ordinary fit-page/fit-width scrolling, and Flutter UI continue through Canvas/Impeller.

That distinction is intentional: Flutter Canvas is already GPU-composited through Impeller, while `package:flutter_gpu` is a low-level API requiring us to implement PDF rendering primitives and compositing ourselves. The direct backend must therefore prove both broad correctness coverage and a material performance win before its scope is expanded or Canvas fallback is removed.

Two immediate problems were found during a side-by-side performance investigation:

1. The checked-in shader bundle is incompatible with Flutter 3.47.
2. Representative CAD pages currently have a 0% direct-GPU acceptance rate.

## Current architecture

- `AppDevTools.tileRasterBackendMode` selects `flutterGpu` by default on supported native platforms.
- `FlutterGpuTileRasterBackend` is used lazily by the deep-zoom tile pyramid.
- A rejected or failed scene permanently uses `PdfCanvasTileRasterBackend` for that scene.
- Initial page rendering and ordinary document scrolling do not use the custom GPU backend.
- Web intentionally uses the compile-time Canvas stub.
- Exact rendering remains the priority: unsupported PDF constructs must fall back rather than be approximated silently.

Relevant code:

- `app/lib/devtools.dart`
- `packages/dart_pdf_editor/lib/src/pdf_page_view.dart`
- `packages/dart_pdf_editor/lib/src/tile_raster_backend.dart`
- `packages/dart_pdf_editor_flutter_gpu/lib/src/backend_native.dart`

## Finding 1: Flutter 3.47 shader bundle is stale

The Dart API compatibility work landed in `e7f7589d` and the repo moved to Flutter 3.47 in `4435410b`, but `assets/shaders/pdf_tile_gpu.shaderbundle` was not rebuilt with Flutter 3.47's `impellerc`.

A native accepted-scene corpus run fails with:

```text
Unsupported shader bundle format version: 1, expected: 2.
This shader bundle was compiled with an incompatible version of impellerc.
Please rebuild the shader bundle with the version of impellerc that ships
with the current Flutter SDK.

[...]

PR #758, opened 2026-08-24
Retain destination-dependent PDF blends on Flutter GPU

ben-milanko:
## Headline

- **Validation outcome: keep this PR in draft; the current ordered-submit route is not a production performance improvement.**
- **Six-run Metal first tile:** **208 ms on `main` -> 222 ms on this PR (+6.7%)** for the real destination-blend fixture.
- **Exact Canvas on the PR runner:** **9 ms p50** for the same fixture; a local warm check was also slower on GPU (**11.7 ms vs 5.3 ms Canvas**).
- **Corpus count:** Ghent **35/22 -> 37/20**, but GWG220/GWG221 contain blend-state markers rather than selected advanced-blend paints, so this does not validate the shader's runtime value.
- **Correctness and safety remain sound:** mean channel difference **0.005/255**, refreshed Metal corpus **218/218**, and dense tiles fall back before compilation.

<details>
<summary>Implementation and validation details</summary>

### What changed

- Implements all destination-dependent PDF blend functions in the Flutter GPU shader: Overlay, Darken, Lighten, ColorDodge, ColorBurn, HardLight, SoftLight, Difference, Exclusion, Hue, Saturation, Color, and Luminosity.
- Preserves premultiplied source and destination alpha with the PDF/W3C compositing equation.
- Renders advanced blend sources into a transparent target, composites them against the accumulated destination, and preserves ordinary painter-order spans between blends.
- Uses ordered command-buffer submissions. Flutter 3.47 Metal can crash when a pass samples a texture written earlier in the same command buffer.
- Keeps Multiply and Screen on their cheaper exact fixed-function path.
- Bounds the ordered route to three destination-dependent paints per tile. A denser tile uses exact retained-scene Canvas replay before GPU scene compilation; sparse tiles on the same page can still use GPU.
- Skips proactive scene compilation when the full scene exceeds the same bound.
- Covers both MSAA and non-MSAA operation; shader copies are required because attachment load after copy does not preserve shader-readable offscreen contents on Flutter 3.47 Metal.
- Adds a dedicated diagnostics counter for bounded blend tile fallbacks.

### Validation

- `fvm dart analyze`: no issues.
- Full Metal-enabled package suite: **250 passed, 3 expected skips**.
- Full refreshed corpus: **218/218 passed**.
- Corpus acceptance:
  - Ghent: **35 accepted / 22 fallback -> 37 / 20**
  - PDF.js: **107 / 72 -> 107 / 72**
- All 13 advanced modes compare against Canvas in painter order with and without MSAA when the explicit test limit is raised:
  - whole-image mean channel difference below **2/255**
  - sampled interior maximum at most **12/255**
- The default-bound 13-mode stress tile skips scene warm-up, performs zero GPU submissions, and matches Canvas byte-for-byte.
- Low-level shader-readable render-target probes cover ColorBurn and HardLight across ordered queue submissions.

### Why it is not merge-ready

The implementation needs a single-submit destination-read route (or another way to remove per-blend ordered submissions) before enabling it in production. The exact Canvas fallback is currently both simpler and faster for the exercised advanced-blend page.

</details>

Experimental work for #682. PR #760 carries the independently worthwhile identity-group improvement.


ben-milanko:
Closing as a measured no-ship experiment. Exact destination-dependent blending requires extra passes/submissions; repeated same-host tests were slower than the existing Canvas fallback, with no coverage or fidelity advantage. The retained single-pass mask/group work continues in #760–#763, while ColorBurn/HardLight remain conservatively routed to Canvas.

PR #761, opened 2026-08-25
Retain GWG vector soft-mask stacks on Flutter GPU

ben-milanko:
### Performance compared with main\n\n- GWG168 first tile p50: 598.8 ms → 514.1 ms (-14.2%); warm retained GPU: 10.8 ms vs 16.2 ms Canvas (-33.5%).\n- GWG169 first tile p50: 577.2 ms → 451.1 ms (-21.8%); warm retained GPU: 10.0 ms vs 14.9 ms Canvas (-33.1%).\n- Ghent GPU acceptance: 35/57 → 37/57. PDF.js stays 107/179.\n\nThe six-run measurements used one macOS Metal host. CI now repeats GWG168 and GWG169 six times against the exact PR base with balanced run order.\n\n<details>\n<summary>Implementation and validation details</summary>\n\nThis retains two exact soft-mask shapes that occur in the refreshed GWG corpus:\n\n- an opaque Normal vector base followed by nested image-masked fills;\n- an axial gradient mask applied to a retained vector fill.\n\nThe lowering is conservative. Partial bases, non-binary outer masks, unsupported transfer/backdrop values, radial masks, non-Normal outer composites, text, and general nested groups still fall back to Canvas.\n\nThe GWG comparison exposed a shared Canvas bug: a mask form inherited the captured source final blend mode. Resetting the mask form to Normal fixes the bevel/emboss actual objects in GWG1610 and GWG168 so they match the reference objects printed on those pages. Only those two baselines and their composite-suite page changed, and each diff was visually reviewed.\n\nValidation:\n\n- full Metal GPU package: 250 passed, 3 intentionally skipped;\n- combined GPU corpus: 218/218 completed;\n- refreshed Ghent render baselines: 54/54;\n- PDF.js Flutter render smoke: 164/164;\n- focused Canvas soft-mask regression: pass;\n- full workspace analyzer: clean.\n\nThe branch is independent of PR #760, but both touch the GPU compiler and this PR should be rebased if #760 lands first.\n</details>

PR #762, opened 2026-08-25
Retain outlined text soft masks on Flutter GPU

ben-milanko:
## Headline

Against the refreshed Ghent Working Group corpus, the new route cuts the local warm tile time from 20.201ms to 12.938ms on GWG1610 (-36.0%) and from 17.126ms to 10.272ms on GWG1611 (-40.0%). Ghent GPU coverage moves from 37 accepted / 20 fallback pages to 39 / 18; PDF.js remains exactly 107 / 72.

This is stacked on #761 so its diff and CI A/B isolate the text-mask work. After #761 merges I will rebase it onto main.

## What changed

- retain outlined fill-only text under image and axial-gradient soft masks
- flatten the exact GWG inner-shadow/glow shape: an opaque text base, one nested image-masked fill, and an identical opaque white text mask
- extend the same proven flattening route to matched non-rectangular opaque path masks
- keep strokes, partial masks, mismatched geometry, multi-fill overprint, and unsupported blend shapes on Canvas
- benchmark both refreshed GWG1610 and GWG1611 scenarios with six baseline and six PR runs in CI

<details>

- `fvm flutter test --enable-impeller --enable-flutter-gpu`: 253 passed, 3 intentional skips
- refreshed Ghent corpus: 39 accepted / 18 rejected
- PDF.js GPU corpus: 107 accepted / 72 rejected
- `fvm dart analyze`: clean
- `git diff --check`: clean
- local 512px visual parity: GWG1610 mean absolute channel difference 2.211/255; GWG1611 1.156/255

The remaining Ghent fallbacks are 11 non-black overprint pages, 3 grouped-text pages, ColorBurn, HardLight, one empty group clip, and one non-identity group.

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -237,6 +237,8 @@ jobs:
             if [[ "$label" == "vector-mask" ||
                   "$label" == "vector-stack-1" ||
                   "$label" == "vector-stack-2" ||
+                  "$label" == "text-mask-1" ||
+                  "$label" == "text-mask-2" ||
                   "$label" == "hairline" ]]; then
               route_change=1
             fi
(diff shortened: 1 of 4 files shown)


ben-milanko:
Ready to merge after #761: all checks pass against the refreshed Ghent corpus. Same-host first-tile p50 improves 746→595 ms (-20.2%) on GWG1610 and 883→691 ms on GWG1611 (the latter overlaps run spread); warm local replay is 36–40% faster than Canvas.

<details>
<summary>Why the cold number is not the visible first page</summary>

The benchmark deliberately calls a cold GPU session directly. In the viewer, the normal full-page raster lands first, then `PdfPageView` creates and warms the scene session only after useful pixels and an idle window; detail tiles await that same warm-up future. The cold geometry/pipeline work is therefore prepared behind the already-visible page rather than replacing its initial raster. Global CI signals are also healthy: raster p95 -15.0%, jank total p95 -2.1%, and unrelated scenario p50s remain within run spread.

Validation: 253 Metal tests passed (3 intentional skips), Ghent 39/18, PDF.js 107/72, analyzer clean.

</details>

PR #763, opened 2026-08-25
Skip definitively failed images on Flutter GPU

ben-milanko:
## Headline

PDF.js GPU coverage increases from 107 accepted / 72 fallback pages to 108 / 71 with exact Canvas parity; refreshed Ghent remains 39 / 18. A definitively failed image decode now stays the same transparent no-op Canvas already renders instead of forcing the whole page off Flutter GPU.

This is stacked on #762 and will be rebased after the prerequisite GPU PRs merge.

## What changed

- record whether a retained scene actually attempted image decoding
- compile a decoder-attempted missing image as a no-op
- continue rejecting deliberate vector-only progressive scenes, where image pixels may arrive in a replacement scene
- cover both states plus pixel parity in a Metal regression test

<details>

- `fvm flutter test --enable-impeller --enable-flutter-gpu`: 254 passed, 3 intentional skips
- refreshed Ghent corpus: 39 accepted / 18 rejected
- PDF.js GPU corpus: 108 accepted / 71 rejected
- `fvm dart analyze`: clean
- `git diff --check`: clean

</details>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -237,6 +237,8 @@ jobs:
             if [[ "$label" == "vector-mask" ||
                   "$label" == "vector-stack-1" ||
                   "$label" == "vector-stack-2" ||
+                  "$label" == "text-mask-1" ||
+                  "$label" == "text-mask-2" ||
                   "$label" == "hairline" ]]; then
               route_change=1
             fi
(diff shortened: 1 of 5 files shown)


ben-milanko:
Ready: all checks are green; this removes one false Canvas fallback without changing the successful-scene runtime path.

<details>

- Refreshed Ghent suite: 39 GPU-accepted / 18 exact fallbacks, unchanged from the stack base.
- PDF.js suite: 108 accepted / 71 exact fallbacks, one additional exact GPU page.
- Local package suite: 254 passed / 3 skipped; analyzer clean.
- CI: test, Flutter GPU smoke/perf, macOS/Linux/Windows GPU, Android, iOS Simulator, web/build/deploy, and both Patrol reports all green.
- Same-host GPU scenario medians stayed within ordinary run spread. Aggregate UI/raster p95 moved +2.4%/+2.6%, with no structural or execution-path change for already successful scenes, so this reads as tail noise rather than a regression.

The new no-op applies only after a definitive image decode failure and only when the remaining progressive scene is otherwise exact; vector-only progressive images still reject to Canvas.

</details>

PR #822, opened 2026-08-26
feat(gpu): retain advanced blends inside groups

ben-milanko:
## Performance first

Compared with the parent commit on the unchanged mixed group and blend workload, repeated same-host medians are:

- settled GPU: 19.95 ms vs 20.27 ms, 1.6% faster
- issue latency: 0.98 ms vs 1.04 ms, 5.5% faster
- cold path: 432 ms vs 422 ms, 2.4% slower and within observed startup noise

The new destination-sampled internal-group workload records 0.52 ms issue latency and 13.4 ms settled GPU time versus 12.1 ms Canvas time on local SwiftShader. That is an explained 11% completion tradeoff for moving exact work off the caller; native Patrol performance is the merge gate.

## What changed

- retain all supported PDF blend modes inside isolated transparency groups
- reproduce Canvas darken for unresolved colored fill and stroke overprint inside those groups
- keep top-level unresolved overprint on Canvas after Ghent showed that broad darken substitution is unsafe
- ping-pong bounded group resolves and destination-sample each non-Normal paint exactly
- preserve distinct per-paint arbitrary clips
- account for the larger transient allocation and fall back before the route budget is exceeded
- add parity tests and a dedicated future performance telemetry case

<details>
<summary>Correctness and corpus evidence</summary>

Focused Canvas parity stays below 4 mean byte difference for internal Overlay, Multiply, Screen, image, text, clip, and colored-overprint cases.

Full corpus totals remain:
- Ghent: 49 accepted / 8 rejected
- PDF.js with system fonts: 177 accepted / 2 rejected

The Ghent ICC-CMS page now advances past the group path-clip gate, then correctly remains on Canvas for unrelated top-level non-black overprint. A controlled comparison improved page 0 from 9.94 to 8.48 mean difference when only the internal group changed from source-over to exact darken. Applying darken to top-level overprint worsened it to 18.08 and exposed incorrect large X marks, so that expansion was deliberately rejected.

</details>

<details>
<summary>Validation</summary>

- fvm dart analyze packages/dart_pdf_editor_flutter_gpu
- complete dart_pdf_editor_flutter_gpu suite with Impeller and flutter_gpu: 308 passed, 4 environment skips
- focused backend suite: 75 passed, 2 environment skips
- full Ghent and PDF.js GPU corpus sweeps
- three repeated same-host control benchmarks
- three isolated internal-group benchmark samples

</details>


ben-milanko:
Reopening immediately to retrigger required workflows on the rebased head; the previous checks are attached to an obsolete SHA.

PR #855, opened 2026-08-27
fix(cos): a sparse progressive buffer's holes are misses, not scans

ben-milanko:
## The freeze

A 219 MB, 83-page pack froze the macOS app for **11.6 s** on open. A profile-mode
run with `--dart-define=PDF_PERF_LOG=true` named it in one line:

```
[perf 12675] JANK build=11619.1ms raster=7.3ms total=11629.5ms
```

One frame, all of it in *build*, no interpret or raster inside it. A CPU profile
from the VM service gave the stack:

```
RenderSliverFixedExtentBoxAdaptor.performLayout
  SliverChildBuilderDelegate.build
    _PdfViewerState._formFieldRects → PdfPage.annotations
      CosDocument.getObject → CosParser.parseIndirectObject → CosLexer.nextToken
        CosLexer.skipWhitespaceAndComments   ← 8515 of 12873 samples
```

### Cause

`PdfSourceLoadOptions(firstPaintPages: 1)` — the app's progressive open —
assembles a **sparse** buffer: header, xref chain and the first page's render
closure are fetched; everything else is left as zeros, on the loader's
documented assumption that they are "zeros the parser never reads".

The viewer reads them. Building a page tile calls `_formFieldRects(index)` →
`PdfPage.annotations`, and annotations are in no page's first-paint closure. And
**0x00 is PDF whitespace**, so a parse starting in a hole skips forward until the
next populated byte — hundreds of megabytes, per object, during layout. The
failed parses then trigger the `N G obj` rescue scan across the whole 219 MB
buffer.

Headless repro (the app's own options, reading what the sliver reads):

| page | before | after |
|---|---|---|
| 0 | 218 ms | 32 ms |
| 1 | **13 213 ms** (answering 0 annotations, not 65) | **0 ms** |
| 4 | 573 ms | 0 ms |

12-page sweep: ~16 s → 34 ms, same answers.

### Fix

`_ProgressiveLoader` already merges the ranges it fetched. They now travel into
`CosDocument.open(populatedRanges:)`:

- an offset inside a hole makes the reference dangle — exactly as a
  not-yet-fetched compressed object already did;
- a parse is bounded to the populated run it starts in (`Uint8List.sublistView`,
  a view, not a copy), so an object at the tail of a fetched range cannot walk
  into the hole after it;
- `_scanObjectHeaders` skips holes, so the rescue scan costs the fetched bytes
  (7 MB here) instead of the file (219 MB);
- an appended incremental revision extends the map, so an update can never read
  as a hole.

### …and the map has to follow the buffer

That fixed the repro and changed nothing in the app — still an 11.5 s frame, same
stack. A `cos open` diagnostic line (kept; it earns its place) showed why:

```
[perf 538] cos open bytes=219314443 sparse spans=110    ← the preview document
[perf 558] cos open bytes=219314443 whole-file          ← what the viewer reads
```

A preview tab paints through `PdfReader(bytes: previewBytes)`, and `PdfReader`
opens **its own session** over those bytes (`PdfShellSessionLifecycle` →
`PdfEditingController` → `CosDocument.open`). The sparse document is used for
nothing but its buffer, and a `Uint8List` cannot carry which parts of it are
real.

So sparseness now belongs to the buffer: `cosSparseBufferRanges`, an `Expando`
the loader stamps and `CosDocument.open` consults when no ranges are passed.
Every later open of that exact buffer inherits the map — no signature change in
the four widgets between the loader and the parser.

**Measured in the app, same document, same machine:**

| | before | after |
|---|---|---|
| worst frame during open | 11 619 ms | 86 ms |
| page 1 on screen | ~10.4 s | 0.6 s |

No viewer change was needed: with holes cheap the layout-time read is ~0 ms, and
the annotation caches are already cleared on the document swap. The render worker
still opens its own *copy* whole-file — harmless here (it records page content,
exactly what the closure holds) and off the UI thread; the init message is where
ranges would go if that ever shows up.

## Also in this branch

- **GPU diagnostics.** The per-page GPU/Canvas route overlay
  (`pdfDebugShowGpuRasterRoutes`, #853) was only reachable from the example app;
  it now has a switch in the app's F12 panel, persisted with the other devtools
  options. `PdfPerfLog` gains `tile gpu compile` (the per-scene GPU cost that
  used to show only as a gap), `tile gpu raster fallback` (a scene that was
  *accepted* and then fell back at raster time — previously invisible in a
  trace), and pipeline/scene warm completion + failure lines.
  `FlutterGpuTileBackendStats.logPerfSummary` brackets every trace
  (`trace-start`/`trace-end`) and each backend switch with lifetime totals.
- **Documents open in Hand mode**, so a mouse drag pans instead of starting a
  text selection.
- **Middle-button drag pans.** Flutter's drag recognizers are primary-only, so
  the gesture reached nothing at all before. A middle drag now grabs the page
  past whatever the primary button would have meant there, without disarming the
  armed tool.

## Tests

New: `pdf_cos/test/byte_source_test.dart` (a hole dangles instead of scanning; a
hole stays authoritative even when real bytes sit inside it; a parse cannot run
off the end of its range; an appended revision is not another hole),
`middle_button_pan_test.dart` (both cases fail without the recognizer change),
`hand_mode_default_test.dart`, and the GPU summary-line tests.

Green locally: `pdf_cos`, `pdf_document`, `pdf_graphics`, `dart_pdf_editor`,
`dart_pdf_editor_flutter_gpu`, `app`, plus `dart analyze` over the workspace.

### DIFF
--- a/app/lib/devtools.dart
+++ b/app/lib/devtools.dart
@@ -196,6 +196,8 @@ class AppDevTools extends ChangeNotifier {
       (PdfPageView.debugTileStoreOverride ?? PdfTileStore.instanceOrNull)
           ?.invalidate();
     }
+    _flutterGpuTileRasterBackend?.stats
+        .logPerfSummary('backend-switch-to-${mode.name}');
     addLog('devtools: tile raster backend → ${mode.name}');
   }
 
@@ -407,6 +409,12 @@ class AppDevTools extends ChangeNotifier {
 
   set logPerfTrace(bool value) {
     if (_logPerfTrace == value) return;
+    // Bracket the trace with the GPU backend's lifetime totals: turning the
+    // log ON stamps the baseline the following lines accumulate from, and
+    // turning it OFF stamps where they ended, so an exported trace carries
+    // both without the reader having to open the panel and read counters that
+    // have since moved. Emitted while the log is still on in both directions.
+    if (!value) _flutterGpuTileRasterBackend?.stats.logPerfSummary('trace-end');
     _logPerfTrace = value;
     PdfPerfLog.enabled = value;
     // The trace is verbose - a line per worker request/reply, interpret, and
@@ -417,6 +425,9 @@ class AppDevTools extends ChangeNotifier {
         ? (m) => _record(DevLogLevel.info, 'perf: $m', notify: false)
         : null;
     addLog('devtools: perf trace logging ${value ? 'on' : 'off'}');
+    if (value) {
+      _flutterGpuTileRasterBackend?.stats.logPerfSummary('trace-start');
+    }
     _scheduleNotify();
   }
 
@@ -669,6 +680,9 @@ class AppDevTools extends ChangeNotifier {
       if (map['renderWindow'] case final bool v) {
         pdfDebugShowRenderWindow.value = v;
       }
+      if (map['gpuRouteOverlay'] case final bool v) {
+        pdfDebugShowGpuRasterRoutes.value = v;
+      }
       if (map['perfOverlay'] case final bool v) {
         showPerformanceOverlay.value = v;
       }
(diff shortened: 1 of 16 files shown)


ben-milanko:
Remaining follow-up tracked in #857: the hole map is keyed on the buffer, so the render worker's and the overprint-retry isolate's *copies* still open whole-file. Latent rather than live — both are off the UI thread and record content the first-paint closure already holds.

PR #859, opened 2026-08-28
fix(editor): unfreeze the eyedropper, reach every page, offer it in the colour dialog

ben-milanko:
## Summary

Four complaints about the colour picker, which turned out to be four separate defects in the same tool.

**1. Arming the eyedropper froze the app.** `_EditingPageOverlayState.build` warmed the sampling raster for whichever page it was building — and the overlay is mounted for every page in the lazy list's cache extent, so arming the tool kicked off one **full page render per mounted page at once**. Each went through `PdfPageColorSampler.of`, which ran the content-stream parse and interpreter walk on the **UI isolate**, then a rasterize and a `toByteData`. It also passed no `maxImagePixelRatio`, so every image decoded at native resolution for a raster that is only ever read three pixels at a time.

The warm is now lazy — only the page the pointer actually reaches (`MouseRegion.onEnter`, `_updatePickPreview`) — and `build` does the opposite, releasing the raster when the tool is put away (the overlay often stays mounted for a selection, and a page of RGBA is megabytes). `PdfPageColorSampler.of` now takes the viewer's `PdfRenderWorker` and records through it like page rendering does, leaving only the replay and the raster on the UI thread; a worker that declines still falls back to the local render. `PdfPageColorSampler.maxPixels` (4 MP) caps the raster — an A0 sheet was 32 MB of RGBA — and `colorAt` maps into the scaled raster, so callers still speak page points.

**2. It only worked on the page it was armed over.** The overlay kept its pan recognizer while the eyedropper was armed. `_panStart` does nothing for it (`case null: break; // eyedropper only - taps, no drags`) but the recognizer still *claimed* the drag, so the document could not be scrolled for as long as the tool was armed. The pan/tap/double-tap callbacks are now dropped while picking, the way cropping drops them. Two consequences fell out: the raw `Listener`'s pointer-up is the commit and fires even when the scroll view won the gesture, so ending a touch scroll picked whatever colour the finger lifted over — a touch/stylus pointer past `kTouchSlop` is now marked a drag and commits nothing (mouse/trackpad keep press-drag-release, which is deliberate); and a page scrolled out from under a live pointer still delivers moves and an up to an unmounted overlay, which tripped `ValueNotifier used after being disposed`. The viewer's `_touchPanEnabledAt` also stood the zoomed touch-pan down while picking on the assumption the overlay handled it, so a zoomed page could not be panned at all — the eyedropper is now on the "viewer owns pan" side of that gate.

**3. The preview chip ignored zoom.** It lives inside the viewer's zoom transform like the rest of the overlay, and alone among the chrome did not counter-scale — a billboard at 800%, unreadable at 25%. It now scales by `_chromeScale` (top-left anchored, offset scaled to match). The *sample* is zoom-aware too: `PdfPageColorSampler.patchRadiusForZoom` sizes the patch in screen pixels — 3×3 at 1:1, a single pixel past 2×, wider zoomed out — instead of always averaging a 3pt disc, which at high zoom is most of a glyph stem.

**4. Sampling from the colour dialog.** `PdfEditingController.pickColorFromPage()` arms the eyedropper and resolves with the sample, and unlike `startColorPick` does *not* adopt it as the annotation colour — the caller asked for a colour of its own. `PdfColorPicker` grows an `onPickFromPage` eyedropper button; the picker can't sample the page it is covering, so `showPdfColorPicker` closes the dialog after firing it and `pickEditingColor` loops: close, arm, wait for the tap, reopen seeded with the sample. Since all editing chrome already routes through `pickEditingColor`, that puts the eyedropper everywhere a colour is chosen. Pickers nested in another modal pass `fromPage: false` (the signature dialog is the one stock case).

Notes: [doc/dev-log/2026-08-28-eyedropper-freeze-reach-dialog.md](doc/dev-log/2026-08-28-eyedropper-freeze-reach-dialog.md).

### API impact

- `EditingPageOverlay(renderWorker:)` — new optional parameter.
- `PdfPageColorSampler.of(worker:, pageIndex:)`, `colorAt(radius:)`, `patchRadiusForZoom`, `maxPixels` — additive.
- `PdfColorPicker(onPickFromPage:)` / `showPdfColorPicker(onPickFromPage:)` — additive, null keeps today's behaviour.
- `PdfEditingController.pickColorFromPage()` — new.
- `buildMultiPagePdf(width:, height:)` — additive, defaults to US Letter.

`finishColorPick` no longer sets `color` when a `pickColorFromPage` caller is waiting; with no such caller it behaves exactly as before.

## Affected packages

- [ ] `pdf_cos`
- [ ] `pdf_document`
- [ ] `pdf_graphics`
- [x] `dart_pdf_editor`
- [x] `pdf_test_fixtures`
- [ ] Example app

## Change type

- [x] Bug fix
- [x] Feature
- [ ] Rendering change
- [x] Editing behavior change
- [x] Performance change
- [ ] Refactor / cleanup
- [ ] Tests / fixtures only
- [ ] Documentation only

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze`
- [ ] Example app smoke test

Full `dart_pdf_editor` suite: 2486 passed, 27 skipped, 0 failed. `dart analyze` clean across `dart_pdf_editor` and `pdf_test_fixtures`. The three pure-Dart packages and the corpus/Ghent suites were not run: nothing here touches the parser, the COS layer, or page rendering — the only rendering-adjacent change is on the eyedropper's own sampling raster, which no baseline covers.

New tests:

- the eyedropper samples a page other than the first;
- a touch drag scrolls while the eyedropper is armed, and its lift is not a sample;
- the chip keeps its size when the page is zoomed;
- `patchRadiusForZoom` across zooms; an oversized page samples through a scaled-down raster;
- `pickColorFromPage` hands the sample back instead of adopting it, and resolves null on cancel;
- the dialog's eyedropper button shows only when the host offers it, closes the dialog, and reopens on the sample; a cancelled sample ends the whole choice.

## User experience

- [x] Completion, cancellation, disabled, loading, and error states were considered.
- [ ] Any journey-level latency, frame-time, or memory impact was measured or ruled out.

On cancellation: Escape and the toolbar button both resolve a waiting `pickColorFromPage` with null, which ends the dialog flow rather than bouncing the picker back at the user; `dispose` resolves it too, so a closed tab can't strand an awaiting caller. Loading: the chip shows `…` until the raster lands, unchanged. Per device — mouse keeps hover preview and press-drag-release; touch and stylus get tap-to-sample with drags going to the scroll view; the zoomed touch-pan path was the specific fix in the viewer.

The perf checkbox is left unticked deliberately. No scenario in `tool/perf/scenarios.json` or `app/tool/perf/scenarios.json` exercises the eyedropper, and the change is a strict reduction in work on that path — N page renders become at most one, the walk moves off the UI isolate, image decodes gain a resolution cap, an unbounded raster gains a 4 MP ceiling, and the raster is released when the tool is put away. Nothing on the ordinary scroll/open/search/edit paths is touched, so the existing scenarios have nothing to say about it. Adding an eyedropper scenario would be worthwhile follow-up if you want a number rather than an argument.

Fixtures are programmatic: `buildMultiPagePdf` gained optional `width`/`height` so the oversized-page raster test can ask for a 4000×4000 pt page.

## Compatibility checklist

`renderer.dart` now imports `render_worker.dart`; no cycle (`render_worker.dart` is Flutter-free and imports nothing from the render side), and `preview_cache.dart` already depends on it the same way.

## Notes for reviewers

- `editing_overlay.dart` had pre-existing formatter drift — running `tool/format.dart` on the untouched file rewrites ~310 lines. That reformatting was reverted so the diff is only the change; the file is still not what the formatter would emit, exactly as on `main`.
- The worker path passes `imagePixelRatio: scale` (≈1), which will usually miss `PdfCachingRenderWorker`'s ratio bucket for a page already recorded at display resolution, so it records afresh. That is fine for the goal here — the walk is off the UI isolate either way — but it is a knob if the extra record ever matters.
- Image overprint and other cases where the sampler and the on-screen device can disagree are unchanged; the sampler renders through the same plan it always did.

---
_Generated by [Claude Code](https://claude.ai/code/session_013snEHgVFmzGCL1WhKiJMz7)_

### DIFF
--- a/doc/dev-log/2026-08-28-eyedropper-freeze-reach-dialog.md
+++ b/doc/dev-log/2026-08-28-eyedropper-freeze-reach-dialog.md
@@ -1,0 +1,117 @@
+# Eyedropper: the freeze, the one-page reach, the zoom, and the dialog
+
+Four complaints about the colour picker, which turned out to be four
+separate defects in the same tool. All of them are in
+`packages/dart_pdf_editor`.
+
+## 1. Arming the eyedropper froze the app
+
+`_EditingPageOverlayState.build` warmed the sampling raster:
+
+```dart
+if (_controller.isPickingColor) unawaited(_ensureSampler());
+```
+
+Every mounted page overlay ran that, and the overlay is mounted for every
+page in the lazy list's cache extent - so arming the tool kicked off one
+**full page render per mounted page, all at once**. Each of those went
+through `PdfPageColorSampler.of`, which called
+`renderPictureRecordedWithPlan` on the **UI isolate**: the content-stream
+parse and interpreter walk, then a rasterize and a `toByteData`, times
+however many pages happened to be mounted. On an ordinary report that is
+the whole freeze; on anything raster-heavy it was worse, because the
+sampler passed no `maxImagePixelRatio` at all, so every image decoded at
+its native resolution for a raster that is only ever read three pixels at
+a time.
+
+Three changes:
+
+- **Lazy, per-page.** The warm moved out of `build` and onto the pointer:
+  `MouseRegion.onEnter` and `_updatePickPreview`. Only the page the
+  pointer actually reaches builds a raster. `build` now does the opposite
+  - `_dropSampler()` when the tool is put away, since the overlay often
+  stays mounted for a selection or another tool and a page of RGBA is
+  megabytes.
+- **Off the UI isolate.** `PdfPageColorSampler.of` takes the viewer's
+  `PdfRenderWorker` (plumbed `PdfViewer` → `EditingPageOverlay` →
+  sampler) and records through it, exactly like page rendering. Only the
+  replay and the raster stay on the UI thread. A worker that declines (an
+  inline image, no worker on this platform) falls back to the local
+  render as before.
+- **A pixel ceiling.** `PdfPageColorSampler.maxPixels` (4 MP) caps the
+  raster; a bigger page is rasterized scaled-down and `colorAt` maps into
+  it, so callers still speak page points. An A0 sheet was 32 MB of RGBA,
+  a large-format plan far more. Images are capped to the raster's own
+  resolution now too.
+
+## 2. It only worked on the page it was armed over
+
+The overlay's `GestureDetector` kept its pan recognizer while the
+eyedropper was armed. `_panStart` does nothing for it (`case null: break;
+// eyedropper only - taps, no drags`) - but the recognizer still *claimed*
+the drag, so the document could not be scrolled for as long as the tool
+was armed. The eyedropper reached exactly the pages that were on screen
+when it was armed, which is what "doesn't work across more than one page"
+was.
+
+`onPanStart/Update/End` (and the double-tap callbacks) are now dropped
+while picking, the same way cropping drops them, so the scroll view wins
+the drag. Two consequences had to be handled:
+
+- The raw `Listener`'s pointer-up is the eyedropper's commit, and it
+  fires whether or not the gesture went to the scroll view - so ending a
+  touch scroll picked whatever colour the finger lifted over and put the
+  tool away. `_pickDragged` marks a touch/stylus pointer that has passed
+  `kTouchSlop`; a dragged pointer's lift commits nothing. Mouse and
+  trackpad keep press-drag-release (those devices don't drag-scroll a
+  list), which is a deliberate feature - you can watch the preview chip
+  while you hunt for the colour.
+- A gesture keeps being routed to the render object it went down on, so a
+  page scrolled out from under a live pointer delivers moves and an up to
+  an *unmounted* overlay - which promptly tripped
+  `ValueNotifier used after being disposed`. The raw handlers and both
+  repaint bumps now tolerate being called late.
+
+The viewer's `_touchPanEnabledAt` also stood the zoomed touch-pan
+recognizer down while picking, on the assumption that the overlay would
+handle those drags. It doesn't, so a zoomed page could not be panned at
+all. The eyedropper is now on the "viewer owns pan" side of that gate.
+
+## 3. The preview chip ignored zoom
+
+The chip lives inside the viewer's zoom transform, like everything else
+in the overlay - and unlike everything else it did not counter-scale, so
+at 800% it was a billboard covering the thing being sampled and at 25% it
+was unreadable. It now scales by `_chromeScale` (anchored top-left, with
+its offset from the pointer scaled to match), the same treatment the
+selection chrome, handles and rotate knob get.
+
+The *sample* is zoom-aware now too. The 3x3 patch exists so an
+anti-aliased stroke reads as its colour, but 3 points of a magnified page
+is most of a glyph stem, so at high zoom the reading was a blur of the
+paper either side of what the pointer was on.
+`PdfPageColorSampler.patchRadiusForZoom` sizes the patch in *screen*
+pixels instead: 3x3 at 1:1, a single pixel past 2x, wider when zoomed out.
+
+## 4. Sampling from the colour dialog
+
+`PdfEditingController.pickColorFromPage()` arms the eyedropper and
+resolves with the sample, and - unlike `startColorPick` - does **not**
+adopt it as the annotation `color`: the caller asked for a colour for its
+own purposes, and quietly repainting the tool with it would be a second
+edit nobody asked for. `finishColorPick` routes to a waiting caller when
+there is one; `cancelColorPick` and `dispose` resolve it null.
+
+`PdfColorPicker` grows an `onPickFromPage` callback, shown as an
+eyedropper button in the value row. The picker can't sample the page
+itself - it is a dialog *over* the page - so `showPdfColorPicker` closes
+the dialog immediately after firing the callback, and `pickEditingColor`
+loops: close, arm, wait for the tap, reopen seeded with the sample so the
+user can confirm or nudge it. A cancelled sample ends the whole choice
+rather than bouncing the dialog back.
+
+Because every piece of editing chrome already goes through
+`pickEditingColor`, that puts the eyedropper everywhere a colour is
+chosen. Pickers nested inside another modal pass `fromPage: false` (the
+signature dialog is the one stock case) - there is no reachable page from
+there, so the button would be a dead end.
--- a/packages/dart_pdf_editor/lib/src/editing/editing_color_pick.dart
+++ b/packages/dart_pdf_editor/lib/src/editing/editing_color_pick.dart
@@ -10,24 +10,47 @@ import 'editing_controller.dart';
 /// is restored from and persisted back to the preferences, and a committed
 /// colour is recorded as a recent.
 ///
+/// The picker also offers the eyedropper, so any colour choice can be
+/// sampled off the page and not just the annotation colour the toolbar's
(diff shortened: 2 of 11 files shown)


ben-milanko:
## On the Patrol perf report

The Patrol comparison on `b444f49` reports every scenario slower (+16% to +29% elapsed p50, jank p95 +32%, worker decode p95 +72%). Since the PR body claims the perf impact was ruled out by argument, that deserves an answer rather than a shrug. I do not think it is this PR, and here is why.

**No Patrol scenario executes any line this PR changes.** The journey never arms an editing tool — `perf_e2e_test.dart` contains no reference to `PdfEditTool`, `.tool =`, `startColorPick`, `selectAnnotation`, or any colour UI. The editing overlay only mounts for an armed tool, a live selection, a pending flash, or the eyedropper, so `editing_overlay.dart` is not merely unexercised here, it is never built. The same goes for the sampler (`PdfPageColorSampler` is reached only from the eyedropper) and for the viewer's touch-pan gate (touch pan only). The one change on a shared path is `renderer.dart` importing `render_worker.dart`, and that adds nothing to any bundle — `pdf_viewer.dart` and `preview_cache.dart` already import it in every build that has a viewer.

**The report contradicts itself as a regression story.** Structural counts are identical where a real slowdown would move them: journey segments 17/17, worker phase requests 101/101, worker request kinds `detail` 3 / `record` 98 on both sides, reconcile modes and thumbnail paths unchanged. Meanwhile the PR side did measurably *less* work — tile replay requests 319 → 267, prefetch batches 24 → 16, prefetch tiles 54 → 46 — and still reports slower wall-clock. And `rounded-rung-pressure`, which records no worker outcomes, no raster, no reconcile and no tile replay at all, is +16.5% elapsed while its jank p95 *improved* 52.6%.

**The harness reports this shape on unrelated PRs.** #855 was a `pdf_cos` progressive-buffer fix that cannot touch web render timing, and its Patrol comparison read `cad-image-deep-zoom` +3.8%, `heavy-document` +6.9%, `jpeg-consumer-decode` +14.6%, `rounded-rung-pressure` +4.4%, `web-worker` +21.5% — every scenario slower, one of them by more than a fifth. #853 ranged -3.2% to +6.8%. This run sits at the high end of that spread, not outside it.

**And the sample size is below the report's own noise guard.** It ran 3 per cohort and says it only labels overlapping interquartile ranges as `within run spread` at 4 or more, so with n=3 it prints raw percentages it would otherwise suppress. The report also states twice that it is advisory and not a CI gate.

So I am reading this as runner noise plus the harness's systematic PR-vs-cached-baseline bias, and not changing the code for it. Being straight about the limits of that claim: I have not run the web harness on this branch myself, so this is an argument from the report's internals and from precedent, not an independent measurement. What would settle it is a re-run — if the same five scenarios come back at similar magnitudes on the new head (`dcd6c5b`, which only adds tests), that is the baseline drifting rather than anything in the diff.

Worth noting separately: no perf scenario covers the eyedropper in either `tool/perf/scenarios.json` or `app/tool/perf/scenarios.json`, which is why the freeze this PR fixes was invisible to CI in the first place. Adding one would be sensible follow-up, and deliberately out of scope here.

---

ben-milanko:
Adding the native Patrol report to the note above, since it settles the question better than my reasoning did.

**The same commit moves the two platforms in opposite directions.** On `b444f49`, Android reads `gpu-native-page-0-canvas-tile` +50.4%, `gpu-native-page-0-reused-tile` +60.6%, `gpu-native-page-0-scene-warm` +16.8%; iOS Simulator reads the same three at **-26.0%, -61.5%, -15.0%**, with `gpu-native-page-0-first-tile` -52.8% and reconcile p95 -92.7%. One diff cannot make a scenario 60% slower on Android and 61% faster on iOS. Whatever this is measuring, it is not the change.

**And the two cohorts are not comparable.** Every native scenario ran `6 / 2` samples (main / PR), and the PR side executed roughly half the journey: segments 17 → 9, perf events 1574 → 1118 on Android and 1430 → 1020 on iOS, tile prefetch batches 132 → 44. The report flags this itself — *"At least one scenario has fewer than three samples; treat its timing delta as provisional."* A truncated two-run cohort against a full six-run baseline will produce large percentages in whichever direction the runner happened to lean.

For completeness, the native journey is in the same position as the web one: `native_perf_e2e_test.dart` has no reference to `PdfEditTool`, `startColorPick`, or any colour UI, and its traces show no thumbnail or worker activity at all — it is a GPU tile journey. None of this PR's code executes in it.

Nothing to change here either. The web comment above still stands, and a re-run on `dcd6c5b` remains the thing that would confirm it.

---

ben-milanko:
The re-run settles it. I said above that a re-run on `dcd6c5b` would be the check; `dcd6c5b` adds **only tests** — not one line of shipped code differs from `b444f49` — and the same five scenarios moved:

| Scenario | `b444f49` | `dcd6c5b` |
| --- | ---: | ---: |
| `cad-image-deep-zoom` | +16.1% | +6.0% |
| `heavy-document` | +28.7% | +6.6% |
| `jpeg-consumer-decode` | +28.6% | +11.8% |
| `rounded-rung-pressure` | +16.5% | +3.9% |
| `web-worker` | +22.4% | +11.2% |

Jank total p95 went +31.9% → +9.3%, and raster p95 +15.6% → **-3.5%**, changing sign. Identical shipped code against the identical cached baseline, and the "regression" shrank by two thirds to three quarters. That is run-to-run variance, as suspected.

The residual +4% to +12% is the same systematic PR-vs-cached-baseline lean visible on unrelated PRs (#855 read +3.8% to +21.5% for a `pdf_cos` change). Nothing here needs a code change, and I have not made one for it.

---

ben-milanko:
Following up on the native report, including the part that goes against me.

**One of my earlier arguments no longer applies.** I criticised the `b444f49` native run for comparing a truncated 2-sample cohort against a 6-sample baseline. This run is a clean 6/6 on every scenario with the full 17-segment journey on both platforms, so that objection is spent. Android still reports `gpu-native-page-0-reused-tile` +53.0%, `gpu-native-page-0-scene-warm` +50.4% and `native-mobile-tiles` +13.7%.

**What still holds, and more cleanly than before.** With cohorts now balanced on both platforms, iOS Simulator reports **every one of its six scenarios "within run spread"** while Android reports three at +13% to +53% — the same diff, the same commit, and only one of the two hosts sees anything. Structural counts are identical across both: tile prefetch batches 132/132, prefetch tiles 252/252, journey segments 17/17, tile slice classes and rung batches unchanged. The PR performs exactly the same work.

**And the numbers move without the code moving.** `dcd6c5b` differs from `b444f49` only in test files — no shipped code at all. Across that no-op, Android's `scene-warm` went +16.8% → +50.4% and `reused-tile` +60.6% → +53.0%, while iOS went from -15% to -61% (all six scenarios) to entirely within-spread. A signal that swings by 33 points on a commit containing no product code is measuring the host, not the diff.

The decisive point is unchanged and independent of all of this: `native_perf_e2e_test.dart` contains no reference to `PdfEditTool`, `startColorPick`, or any colour UI, and its traces show no thumbnail, worker or reconcile activity. The editing overlay never mounts in that journey, so none of this PR's code executes in it.

Still no code change. If the Android emulator lane's variance is worth tightening, that is its own piece of work on the harness rather than something this PR can answer.

---

PR #872, opened 2026-09-03
fix(viewer): Fix text markup hit testing for disjoint highlight quads

SpanManX:
### Text markup hit testing included invisible gaps
Text markup annotations can contain multiple disjoint /QuadPoints, while their /Rect covers the entire bounding area.
Previously, annotation hit testing used only /Rect. As a result:
- Empty gaps between highlighted text runs were treated as part of the annotation.
- The click cursor appeared over non-highlighted content.
- Clicking those gaps could trigger the annotation callback.

image

### For text markup annotations:
- Use /QuadPoints as the authoritative hit-test geometry.
- Fall back to /Rect only when no valid quad geometry is available.
- Preserve the existing /Rect behavior for other annotation types.

image

## Affected packages

- [ ] `pdf_cos`
- [ ] `pdf_document`
- [ ] `pdf_graphics`
- [x] `dart_pdf_editor`
- [ ] `pdf_test_fixtures`
- [ ] Example app

## Change type

- [x] Bug fix
- [ ] Feature
- [ ] Rendering change
- [ ] Editing behavior change
- [ ] Performance change
- [ ] Refactor / cleanup
- [ ] Tests / fixtures only
- [ ] Documentation only

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze`
- [x] `cd packages/pdf_cos && fvm dart test`
- [x] `cd packages/pdf_document && fvm dart test`
- [x] `cd packages/pdf_graphics && fvm dart test`
- [x] Real PDF corpus parse/render smoke test
- [x] Example app smoke test

## User experience

- [ ] The affected user journey and expected outcome are described above.
- [ ] Relevant keyboard, mouse, trackpad, touch, stylus, focus, and accessibility paths were checked.
- [ ] Preview, commit, undo/redo, save, and reopen behavior agree where applicable.
- [ ] Any journey-level latency, frame-time, or memory impact was measured or ruled out.

## Compatibility checklist

- [ ] No `dart:ui` or Flutter imports outside `packages/dart_pdf_editor`.
- [ ] No `dart:io` imports in any `lib/` directory.
- [ ] Layering is preserved: `pdf_cos` <- `pdf_document` <- `pdf_graphics` <- `dart_pdf_editor`.
- [ ] Parsers remain lenient on real-world input and writers remain strict on output.
- [ ] Raw PDF stream bytes stay lazy/raw until decoding is required.
- [ ] Test fixtures use builders/programmatic generation instead of hand-edited byte offsets.

### DIFF
--- a/packages/dart_pdf_editor/lib/src/pdf_viewer.dart
+++ b/packages/dart_pdf_editor/lib/src/pdf_viewer.dart
@@ -5762,7 +5762,17 @@ class _PdfViewerState extends State<PdfViewer>
     final annots = actionsOnly ? _interactiveAnnots(i) : _visibleAnnots(i);
     // later /Annots entries paint on top, so they win the hit test
     for (final annotation in annots.reversed) {
-      if (annotation.rect.contains(x, y)) {
+      // Text markups may contain several disjoint runs. Their bounding
+      // /Rect includes the whitespace between those runs, so using it here
+      // makes the host tap callback and click cursor claim visibly empty
+      // text. Match the editing controller's selection hit test: visible
+      // /QuadPoints are authoritative, with /Rect as the fallback for every
+      // other annotation type (and malformed markups without quads).
+      final quads = annotation.behavior.markupQuads;
+      final hit = quads != null && quads.isNotEmpty
+          ? quads.any((quad) => quad.contains(x, y))
+          : annotation.rect.contains(x, y);
+      if (hit) {
         return (
           pageIndex: i,
           annotation: annotation,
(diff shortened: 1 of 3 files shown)


ben-milanko (changes requested):
One issue to address before merging:

**[P2] Valid rotated markup still falls back to `/Rect`** — `packages/dart_pdf_editor/lib/src/pdf_viewer.dart:5771-5774`

`behavior.markupQuads` deliberately returns `null` unless every quad is axis-aligned. Rotated/skewed highlights—or arrays containing one malformed quad—therefore retain the original false-hit behavior across whitespace. Hit testing should parse general `/QuadPoints` and use point-in-quadrilateral geometry, falling back to `/Rect` only when no usable quads exist. A rotated, disjoint-quad regression test would cover this.

No other actionable findings. The sole red CI check appears unrelated: it is a GPU buffer-reuse benchmark, while the viewer tests passed.

ben-milanko (approved):
The requested QuadPoints fix is applied: hit testing now accepts rotated/skewed quadrilaterals, skips malformed groups independently, and falls back to /Rect only when no usable quad exists. Added a rotated disjoint-quad regression with a malformed group; the full PdfViewer test file and workspace analysis pass.

PR #875, opened 2026-09-04
fix(viewer): avoid blurry text under multi-line highlights

SpanManX:
## Summary

Fixes blurry text beneath multi-line Highlight annotations, most noticeable on Windows at fractional zoom levels such as 165%.

## Root Cause

Multi-line Highlight annotations were previously rendered as a single `ui.Picture` covering the union of their `/QuadPoints`.

Because Highlights use the `Multiply` blend mode, the Windows GPU compositor could create and resample a large intermediate surface covering multiple text lines. At fractional zoom levels, this caused the text within that area to appear blurred.

## Changes

- Render multi-line Highlights as one tightly clipped `ui.Picture` per `/QuadPoints` region.
- Update the annotation appearance cache to support multiple pictures per annotation.
- Preserve the existing single-picture path for:
  - Single-line Highlights
  - Underline, StrikeOut, and Squiggly annotations
  - All other annotation types
- Preserve the original color, opacity, appearance stream, and `Multiply` blending behavior.
- Avoid page raster invalidation and re-rendering.

## Performance Impact

The change only affects multi-line Highlight annotations.

A two-line Highlight now replays two small pictures instead of one large picture. This adds a small number of draw calls but reduces the intermediate compositing area. It does not trigger page raster refreshes, so the expected performance impact is negligible.

## Issue
image

## Result
image

## Affected packages

- [ ] `pdf_cos`
- [ ] `pdf_document`
- [ ] `pdf_graphics`
- [x] `dart_pdf_editor`
- [ ] `pdf_test_fixtures`
- [ ] Example app

## Change type

- [x] Bug fix
- [ ] Feature
- [x] Rendering change
- [ ] Editing behavior change
- [ ] Performance change
- [ ] Refactor / cleanup
- [ ] Tests / fixtures only
- [ ] Documentation only

## Validation

- [ ] `fvm flutter pub get`
- [x] `fvm dart analyze`
- [ ] Example app smoke test

## User experience

## Compatibility checklist

### DIFF
--- a/packages/dart_pdf_editor/lib/src/pdf_viewer.dart
+++ b/packages/dart_pdf_editor/lib/src/pdf_viewer.dart
@@ -1160,6 +1160,16 @@ class PdfViewer extends StatefulWidget {
     int rotation,
   )? debugAnnotationAppearanceRendererOverride;
 
+  /// List-returning counterpart used to exercise one annotation owning several
+  /// tightly bounded appearance pictures. Takes precedence over
+  /// [debugAnnotationAppearanceRendererOverride].
+  @visibleForTesting
+  static Future<List<ui.Picture>> Function(
+    PdfPage page,
+    PdfAnnotation annotation,
+    int rotation,
+  )? debugAnnotationAppearancePicturesRendererOverride;
+
   /// Test observer called immediately before an annotation appearance picture
   /// is disposed. Production ownership and disposal remain unchanged.
   @visibleForTesting
@@ -8639,7 +8649,7 @@ class _AnnotationAppearanceLayerState
   /// The stream half is typed `Object` because `CosStream` is not exported
   /// into this library; it uses identity equality, so the key mixes stream
   /// identity with the [PdfRect]'s value equality - exactly the intent.
-  final Map<(Object, PdfRect), ui.Picture> _cache = {};
+  final Map<(Object, PdfRect), List<ui.Picture>> _cache = {};
   int? _cacheRotation;
   Size? _cacheSize;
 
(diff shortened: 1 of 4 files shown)


ben-milanko (changes requested):
Two issues to address before merging:

**[P2] Overlapping quads are composited multiple times** — `packages/dart_pdf_editor/lib/src/renderer.dart:600-640`

The split accepts overlapping `/QuadPoints`, then replays the complete appearance once per quad. Overlap pixels are therefore darkened repeatedly. A targeted regression produced alpha 192 instead of the legacy 128. Require pairwise-disjoint clips, merge overlaps, or retain the single-picture path when they overlap.

**[P2] Rendering becomes quadratic for long highlights** — `packages/dart_pdf_editor/lib/src/renderer.dart:621-639`

For every one of `q` quads, the code creates a new interpreter, reparses the appearance, and redraws its complete `q`-quad content. A long highlight therefore performs roughly `q²` work and creates `q` pictures in one scheduled render, potentially blocking the UI. Cap/batch splitting or fall back to the legacy path beyond a safe count.

All current CI checks are green.

ben-milanko (approved):
Both requested renderer fixes are applied: overlapping quads retain the legacy single compositing pass, and split replay is capped at eight pairwise-disjoint quads so long highlights cannot become quadratic. Added pixel-identity and long-highlight regressions; the focused suite and workspace analysis pass.

SpanManX:
> Both requested renderer fixes are applied: overlapping quads retain the legacy single compositing pass, and split replay is capped at eight pairwise-disjoint quads so long highlights cannot become quadratic. Added pixel-identity and long-highlight regressions; the focused suite and workspace analysis pass.

Thanks, both issues are resolved.

Issue #649, opened 2026-08-05
Substituted fonts: glyphs drift up to ~4pt inside a run because only the run total is matched


ben-milanko:
Split out of #647. That issue reported the text selection band sitting beside the glyphs. It has two independent halves:

- **range → geometry** (the selection's own math). Fixed in #648: `quadsFor`/`positionNear` now use exact per-character advances.
- **glyph painting** (where the substituted glyphs actually land). This issue. Still open, and now the visible half.

## The defect

For a run with no embedded font program, `CanvasPdfDevice.drawText` shapes the whole string with a substituted system font and applies **one uniform horizontal scale** so the run total matches the PDF's advance:

```dart
final targetWidth = paintRun.width * renderSize;
final scaleX = targetWidth / layout.width;
...
canvas.scale(scaleX / renderSize, -1 / renderSize);
layout.paint(canvas, Offset(0, -layout.baseline));
```

That pins the run's **endpoints** and nothing in between. Interior characters land at the substitute's own advance distribution, not the PDF's. `_composeLayout` has the same property and says so:

> placing per-character layouts at their cumulative natural advances. The total natural width still feeds the same `scaleX` … only the intra-run distribution differs

`_composableRun` gates composition to text with no kernable adjacency and excludes lowercase outright, so ordinary prose takes the whole-run path.

## Magnitude

Page 1 of the example demo document draws a 77-character line as a single `Tj` in non-embedded 12pt Helvetica. Comparing Helvetica AFM cumulative offsets (what the PDF specifies) against a substitute's, after the uniform `scaleX`:

```
run: 77 chars, total 416.20pt, scaleX 0.8833
drift at "annotations." start (char 28): -3.58pt
drift at its end          (char 40): -2.69pt
worst drift 4.39pt at char 22
endpoints: start 0.0000pt, end 0.0000pt
```

Drift is zero at both ends and peaks mid-run — the signature of matching a total instead of a distribution. Nearly 4.4pt at 12pt type is about two thirds of a lowercase glyph.

Measured with DejaVuSans standing in for the substitute; the real one is whatever the platform picks (Roboto on web/Android), so the exact numbers move but the mechanism and rough scale do not.

## How it shows up

[...]

PR #896, opened 2026-09-06
Stop stretching substituted glyphs across their character spacing

ben-milanko:
## Summary

A CAD-exported bridging form drew its table row numbers (`3`, `4`, `5`, `6`) roughly four times too wide - wide and heavy enough to read as a different typeface than the rest of the sheet. The fonts (ArialMT / Arial-BoldMT, none embedded) carry correct `/Widths`; the rows are

```
15.137 Tc 54.96 396.24 Td
( 3)Tj
```

At `/F1 9 Tf` that Tc is 1.68 em of character spacing per glyph - the export's way of reaching the next column - so the pen steps 2.24 em across a digit the font gives 0.556 em.

Exact substituted-glyph placement (#649) scales glyph shapes by `Σ substitute advance ÷ Σ PDF pen step`, and a pen step is a glyph's width **plus the spacing that follows it**. The run therefore taught the device that the document's glyphs were 4x wider than they are. Spacing is a gap to open, never a shape to stretch.

- `PdfTextRun.glyphWidthAt(index, length)` (new, `pdf_graphics/device.dart`) is the pen step less the run's `letterSpacing`, plus the `wordSpacing` after a single-byte space. It now feeds the uniform scale in all three places that copy-fit substituted glyphs: `CanvasPdfDevice._buildPlacedLayout`, `pdfCanvas2dTextLayout` (worker Canvas2D surface) and `FlutterGpuTrueTypeTextOutliner.outline`. Origins keep using the offsets themselves, so the spacing stays in the geometry selection, search and extraction are computed from.
- With the spacing out of the scale, an unspaced shaped word drifts from the offsets by one Tc per character, so the old shape-the-word-or-explode-it-into-characters rule sent every word of a Tc-carrying document to per-character placement. A word is now cut into the longest **pieces** that shape within the existing 0.02 em tolerance of the PDF's offsets: whole words stay whole where the substitute agrees, and each disagreement is spent at a cut instead of accumulating across the word. That is strictly better than the old binary, which exploded a whole word when only its tail drifted.
- The whole-run path (no offset table, or a gradient/stroke run) was never affected - it reproduces Tc/Tw as real tracking through `TextStyle`, so its measured width already includes the spacing it is scaled against.

Rejected on the way, and written up in the dev-log note: giving a shaped word its Tc back through `TextStyle.letterSpacing` / `context.letterSpacing`. It is faster on spaced text, but both engines distribute letter spacing *around* each glyph (Skia paints a lone `A` with `letterSpacing: 100` at x=50), so every part would need a half-spacing compensation for a rule neither engine documents.

## Affected packages

- [ ] `pdf_cos`
- [ ] `pdf_document`
- [x] `pdf_graphics`
- [x] `dart_pdf_editor`
- [ ] `pdf_test_fixtures`
- [ ] Example app
- [x] Documentation / CI / repository metadata only (dev-log note)

Also `dart_pdf_editor_flutter_gpu`, which the template predates.

## Change type

- [x] Bug fix
- [ ] Feature
- [x] Rendering change
- [ ] Editing behavior change
- [x] Performance change
- [ ] Refactor / cleanup
- [ ] Tests / fixtures only
- [ ] Documentation only

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze` (clean at repo root)
- [x] `cd packages/pdf_document && fvm dart test` (990 passed)
- [x] `cd packages/pdf_graphics && fvm dart test` (1064 passed, Ghent + pdf.js corpus tests included)
- [x] `cd packages/dart_pdf_editor && fvm flutter test` (2694 passed)
- [x] Ghent corpus test or baseline update - ran, **no baseline change needed**: the VM render baselines do not go through the substituting canvas device
- [x] Real PDF corpus parse/render smoke test - the reported document rendered before/after through `PdfPageRenderer.renderImage`
- [ ] Example app smoke test

`pdf_cos` is untouched by this change. `dart_pdf_editor_flutter_gpu`'s suite was also run (15 passed, 106 skipped without a GPU).

### Performance

45 lines of substituted prose, DPR 2, best of 15; raster figures are net of a 6.0 ms empty-page raster, `record` is the cold pass with the layout cache cleared, `raster` the warm one:

| Tc (em) | parts | record | raster (net) |
| --- | --- | --- | --- |
| 0 (before, and after) | 506 | 17.8 ms | 7.5 ms |
| 0.005, per-character | 3042 | 10.3 ms | 10.1 ms |
| 0.005, pieces (this PR) | 944 | 21.6 ms | 8.4 ms |
| 0.02, pieces (this PR) | 2861 | 15.8 ms | 9.9 ms |

A document that sets no Tc - most of them - is untouched: same parts, same cost. One that does pays ~15% more text raster than shaping its words whole would, and avoids the ~35% the per-character fallback would have cost. Steady-state paint recurs on every scroll and zoom; the cold shaping pass runs once per unique run.

## User experience

- [x] The affected user journey and expected outcome are described above: non-embedded text set with character spacing (CAD and office exports do this routinely) now paints at its own glyph widths instead of stretched to fill the spacing.
- [ ] Completion, cancellation, disabled, loading, and error states were considered - not applicable, this is a paint-time change with no new states.
- [ ] Keyboard/mouse/touch/accessibility paths - not applicable; selection, search and hit-testing geometry are unchanged, since only the glyph shapes moved.
- [ ] Preview, commit, undo/redo, save, reopen - not applicable, nothing is written to the document.
- [x] Journey-level latency measured - see the table above.

The reporting document is a customer form and is not attached. The behaviour is pinned by programmatic fixtures instead: `substituted_glyph_placement_test.dart` renders `( 3)Tj` under a `15.137 Tc` and asserts the digit's ink is exactly as wide as an unspaced one and starts on the pen position the PDF gives it, plus a 0.5 em Tc case that must leave four separate glyph blocks rather than one solid stretched word. `web_surface_profile_test.dart`, `text_outliner_test.dart` and `interpreter_test.dart` cover the other three surfaces. All four new tests fail on the pre-fix code.

## Compatibility checklist

## Notes for reviewers

- `PdfTextRun.glyphWidthAt` returns null rather than guessing when there is no offset table or the span falls outside it; every call site declines the placed path in that case, falling back to the established whole-run shaping.
- A `/ToUnicode` code that maps to several characters has its advance split evenly by the interpreter, so each piece is charged the spacing once - documented on the helper. It understates a spaced ligature's pieces; the sum is clamped at zero and a run that clamps to nothing falls back to whole-run shaping.
- The `_placedLayout` cache key gained Tc/Tw: they now set the shape scale and the cut points, and two runs can share an offset table without sharing them.
- Background: doc/dev-log/2026-09-06-character-spacing-is-not-shape.md.

https://claude.ai/code/session_01FyJhDde4UD7KMvwfMy9Aja

---
_Generated by [Claude Code](https://claude.ai/code/session_01FyJhDde4UD7KMvwfMy9Aja)_

### DIFF
--- a/doc/dev-log/2026-09-06-character-spacing-is-not-shape.md
+++ b/doc/dev-log/2026-09-06-character-spacing-is-not-shape.md
@@ -1,0 +1,103 @@
+# Character spacing is a gap, not a shape (over-wide table digits)
+
+A bridging-details table from a CAD-exported form (ArialMT/Arial-BoldMT,
+none of them embedded) drew its row numbers - `3`, `4`, `5`, `6` - roughly
+four times too wide, fat enough to read as a different, bolder typeface than
+the rest of the sheet. Everything else on the page was fine.
+
+The content stream says it plainly:
+
+```
+15.137 Tc 54.96 396.24 Td
+( 3)Tj
+```
+
+At `/F1 9 Tf` that Tc is 1.68 em of character spacing per glyph - the export's
+way of reaching the next column - so the pen steps 2.24 em across a digit the
+font gives 0.556 em, after a leading space that steps 1.96 em.
+
+## The defect
+
+Exact substituted-glyph placement (#649) gives a substituted run one uniform
+horizontal glyph scale and puts every origin on the PDF's own pen offset. The
+scale was `Σ substitute advance ÷ Σ PDF pen step`. A pen step is a glyph's
+width *plus the spacing that follows it*, so a run carrying a Tc taught the
+device that the document's glyphs were that much wider than they are, and the
+shapes were stretched to fill the gap: 2.24 ÷ 0.556 = 4.02x on this table.
+
+The rule the code was missing is that **spacing is a gap to open, never a
+shape to stretch**. The uniform scale must weigh the substitute's glyphs
+against the widths the PDF gives *the same glyphs*, and the spacing must be
+left to the offsets, which already carry it.
+
+`PdfTextRun.glyphWidthAt(index, length)` (pdf_graphics `device.dart`) is that
+width: the pen step less the run's `letterSpacing`, plus the `wordSpacing`
+after a single-byte space. It is now the scale input at all three sites that
+copy-fit substituted glyphs:
+
+- `CanvasPdfDevice._buildPlacedLayout` (dart_pdf_editor `canvas_device.dart`)
+- `pdfCanvas2dTextLayout` (`web_surface_profile.dart`, the worker Canvas2D
+  surface)
+- `FlutterGpuTrueTypeTextOutliner.outline` (flutter_gpu `text_outliner.dart`)
+
+The whole-run path (no offset table, or a gradient/stroke run) was never
+affected: it reproduces Tc/Tw as real tracking through `TextStyle`, so its
+layout width already includes the spacing it is scaled against.
+
+## Words became pieces
+
+Correcting the scale alone breaks the other half of #649. A word was shaped
+whole when its interior agreed with the PDF's offsets within 0.02 em, and
+placed character by character otherwise. With the spacing out of the scale, an
+unspaced shaped word drifts from the offsets by one Tc per character, so every
+word of a document that sets *any* appreciable Tc failed that test and blew up
+into per-character `drawParagraph` calls.
+
+So the word/character binary is now a greedy cut: a word is broken into the
+longest **pieces** that can be shaped tight with no character landing more than
+`_placementToleranceEm` from its own offset, each piece drawn at the offset of
+its first character. A word the substitute agrees with is still one piece;
+disagreement - the substitute's own drift, or a Tc the shaped text does not
+carry - is spent at a cut instead of accumulating across the word. This is
+strictly better than the old binary, which exploded a whole word when only its
+tail drifted.
+
+Rejected on the way: giving a shaped word its Tc back through
+`TextStyle.letterSpacing` (Flutter) / `context.letterSpacing` (Canvas2D). It
+is faster, but both engines distribute letter spacing *around* each glyph -
+Skia paints a lone 'A' with `letterSpacing: 100` at x=50, not x=0 - so every
+part would need a half-spacing compensation for a rule neither engine
+documents, and being wrong about it puts a pathologically spaced glyph (this
+table's, at 0.84 em) visibly off its column.
+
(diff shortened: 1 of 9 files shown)


ben-milanko:
The Patrol web comparison above shows +15–29% elapsed across four scenarios. I checked it rather than waving it through, and it is runner noise, not this diff. The evidence:

**This PR cannot execute differently in any of those scenarios.** Every change here is inside the substituted-glyph placement path, and every behavioural difference is gated on a run's `letterSpacing` (Tc) being non-zero. None of the Patrol fixtures set Tc — `grep " Tc"` over `perf_e2e_test.dart`, `cad_strip.dart` and `cad_image_strip.dart` returns nothing. With Tc = 0:

- `PdfTextRun.glyphWidthAt` returns the pen step unchanged, so the shape scale `k` is bit-identical to `main`;
- with the same `k`, the piece cutter cuts exactly where the old `_wordHolds` returned false, so a word that held is still one shaped part. Where a word did not hold, the new code emits *fewer* parts than the old explode-to-characters branch, never more.

**The workloads are the wrong shape for a text change anyway.** `_buildPerfDocument` (heavy-document, web-worker) is 6000 stroked lines per page plus a single 20-character Helvetica run; `cad-image-deep-zoom` and `jpeg-consumer-decode` are image-tile strips. A `drawText` change cannot cost 2.2 s there.

**The signals that moved are ones this diff does not touch.** Worker decode p95 +85% and `jpeg-consumer-decode` decode 1.0 ms → 2.0 ms are image decode. Tile replay p95 is +103% while tile replay *requests* fell 363 → 327 and prefetch batches 28 → 20 — fewer tiles taking longer each is a slower runner, not more work. `rounded-rung-pressure`, the one scenario with no worker or raster activity, is flat at +0.7%.

At 3 samples per cohort the harness prints raw percentages (it only suppresses them as `within run spread` at 4+ samples with overlapping IQRs), and it labels wall-clock advisory rather than a gate.

For the record, the text-path cost that this PR *does* change was measured directly — 45 lines of substituted prose at DPR 2, best of 15, net of a 6.0 ms empty-page raster: a Tc-free document is unchanged (506 parts, 7.5 ms), and a document setting Tc 0.005 em pays 8.4 ms instead of the 10.1 ms the per-character fallback would have cost. Full table in the PR description.

No change pushed for this comment.

---

PR #897, opened 2026-09-07
Add persistent vertical alignment for text fields

DEV-Jaron:
## Summary

Text fields currently choose vertical placement solely from their multiline flag, so callers cannot centre or bottom-align wrapped text. Add `PdfFormTextVerticalAlignment` and optional `verticalAlignment` arguments to `setTextValue` and `setTextFieldStyle`, plus a saved-setting getter, a `legacy` option that clears the preference during the same fill/style operation, and a standalone clear operation.

The preference applies to the complete wrapped block in every widget and survives saved/reopened fills, restyling, resizing and rotation. Omitted preferences preserve legacy rendering. Overfull blocks stay top anchored and clipped; wrapping, horizontal justification and auto-sizing retain their existing policies.

Persistence follows the existing private `DartPdf…` marker convention. The README documents its representation and interoperability limits. All placement math lives in `writePdfTextBox`: form settings map to `PdfTextBoxVAlign.top`, `centerBlock`, and the new `bottomBlock`. Its opt-in `clampVerticalAlign` keeps overfull form blocks top anchored while preserving existing signature-panel placement. Form auto-size retains its conservative fit metric; the bottom-block calculation documents why placement excludes trailing leading.

## Affected packages

- [ ] `pdf_cos`
- [x] `pdf_document`
- [ ] `pdf_graphics`
- [ ] `dart_pdf_editor`
- [ ] `pdf_test_fixtures`
- [ ] Example app

## Change type

- [x] Feature
- [x] Rendering change
- [x] Editing behavior change

## Validation

Used the repository-pinned Flutter 3.47.0 / Dart 3.13.0 SDK directly (equivalent to the FVM commands):

- [x] `flutter pub get`
- [x] `dart analyze --fatal-infos`
- [x] `pdf_cos`: 373 tests passed
- [x] `pdf_document`: 1,039 passed; 1 skipped
- [x] `pdf_graphics`: 1,064 passed
- [x] `dart_pdf_editor`: 2,690 passed; 29 skipped
- [x] Repository formatting wrapper and lockstep dependency check (the two unrelated pre-existing layouts flagged in review were restored after formatting)
- [x] PDF parse/render smoke test
- [ ] Example app smoke test

New regressions cover single/multiline placement across horizontal alignments, legacy defaults, persistence, clearing during fill/style, shared-helper overflow clamping enabled/disabled, unknown/malformed metadata, wrapping and blank lines, overflow and auto-size floor, multiple widgets, rotation, embedded fonts, read-only/type guards and flattening.

## User experience

This is a library API change; no editor control is introduced. A downstream application's unchanged form-filling function was exercised against the modified library, including its forced `multiline: true` call. Its full app UI was not launched.

Automated regressions use programmatically generated fixtures. Validation before the shared-helper refactor also used a two-page form with fabricated values: all 119 fields and 123 widgets survived interactive fills, with unchanged names, rectangles and page content streams. In Poppler, explicit top matched legacy rendering, and centred interactive and flattened outputs matched pixel-for-pixel. A synthetic alignment comparison was also rendered and visually checked with the library's Flutter renderer. Local form files are not included in this contribution.

## Compatibility checklist

- [x] Layering is preserved.
- [x] Unknown/malformed preference values fall back without throwing.
- [x] Raw PDF stream bytes remain lazy until decoding is required.
- [x] Test fixtures are generated rather than hand-editing byte offsets.

## Compatibility notes

`/DartPdfTextVerticalAlignment` is a library-private text string (`top`, `center`, `bottom`) stored on the terminal field, not inherited; it is not a standard AcroForm property, and no developer-prefix registration is claimed. Other editors may change placement or discard the preference when rebuilding appearances.

Explicit centre aligns the em-height block even for one line; legacy single-line placement remains ascent-centred when no preference is saved. Alignment is applied by the shared helper after existing wrapping/auto-size work. `legacy` removes the private key rather than storing a fourth value; the getter then returns null. No package versions or dependency constraints changed.

### DIFF
--- a/packages/pdf_document/README.md
+++ b/packages/pdf_document/README.md
@@ -57,6 +57,52 @@ editor.addFreeText(0, const PdfRect(72, 600, 280, 660), 'Reviewed.');
 final saved = editor.save(); // incremental update
 ```
 
+## Text-field vertical alignment
+
+Vertical placement is independent of wrapping and horizontal `align` / `/Q`:
+
+```dart
+editor.setTextValue(field, 'First line\nSecond line',
+    multiline: true,
+    verticalAlignment: PdfFormTextVerticalAlignment.center);
+
+// Save a template preference without changing its value.
+editor.setTextFieldStyle(field,
+    verticalAlignment: PdfFormTextVerticalAlignment.bottom);
+
+// Omitted/null alignment retains the preference, including after reopening.
+editor.setTextValue(field, 'A later value');
+print(field.textVerticalAlignment);
+editor.clearTextFieldVerticalAlignment(field); // restore legacy placement
+```
+
+`top`, `center`, and `bottom` position the whole block after wrapping and
+font auto-sizing, inside the existing 2-point padding. Each line occupies an
+em-height box with the existing 1.15-times-font-size baseline spacing; there
+is no extra leading before the first or after the last line. If the block
+cannot fit, it is top anchored and clipped so its beginning stays visible.
+Existing font sizes, auto-size limits and long-word clipping are unchanged.
+With no saved preference, multiline fields retain top placement and single-line
+fields retain their existing ascent-centred placement. Explicit `center`
+centres the em-height block even for a single line, so its baseline can differ
+slightly from legacy single-line placement.
+
+The preference is a **dart-pdf private extension**, following the library's
+existing `DartPdf…` metadata convention. It is stored as a PDF text string
+(`top`, `center`, or `bottom`) in `/DartPdfTextVerticalAlignment` on the terminal
+text-field dictionary, applies to all that field's widgets, and is not
+inherited from parent fields. Absent, unknown or malformed values use legacy
+placement; ordinary fills preserve unknown metadata. Clearing removes the
+entry. Changing value, style, size or rotation regenerates appearances using
+the saved preference.
+
+This key is not a standard PDF property or a claim of a registered developer
+prefix. `/Q` remains horizontal justification. Standard `/AP` appearances
+carry the visible result for other viewers, but editors that do not understand
+the private preference can replace the placement when editing or regenerating
+appearances, and may discard the metadata. A flattened export retains the
+painted placement but no longer has editable fields.
+
 ## Reduce file size
 
 ```dart
--- a/packages/pdf_document/README.md
+++ b/packages/pdf_document/README.md
@@ -73,7 +73,11 @@ editor.setTextFieldStyle(field,
 // Omitted/null alignment retains the preference, including after reopening.
 editor.setTextValue(field, 'A later value');
 print(field.textVerticalAlignment);
-editor.clearTextFieldVerticalAlignment(field); // restore legacy placement
+// Clear the preference while restyling, with one appearance regeneration.
+editor.setTextFieldStyle(field, color: 0x000000,
+    verticalAlignment: PdfFormTextVerticalAlignment.legacy);
+// Or clear it alone:
+editor.clearTextFieldVerticalAlignment(field);
 ```
 
 `top`, `center`, and `bottom` position the whole block after wrapping and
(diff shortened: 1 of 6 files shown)


ben-milanko:
Thanks for this — the feature is worth having and the test coverage is thorough. One structural change I'd like before it lands, plus some smaller notes.

## On the private key: agreed, there's no built-in alternative

To save the next reader the research: there is no standard PDF entry for this. A variable-text field (§12.7.3.3) carries `/Q`, which is horizontal quadding only. Vertical placement is a viewer convention rather than a stored property — Acrobat centres single-line fields and top-anchors multiline ones, which is exactly the `multiline ? top : centerLine` rule `form_editor.dart` already hard-codes. The near-misses that look promising and aren't:

- `/MK` → `/IF` → `/A` is a genuine vertical-alignment fraction, but it's the icon-fit dictionary and applies only to a pushbutton widget's icon.
- Rich text (`/RV`, §12.7.3.4) has a `vertical-align` attribute, but it is a per-span baseline shift (sub/superscript), not block alignment.
- XFA `<para vAlign="middle">` does what we want, and is deprecated in PDF 2.0 and irrelevant to AcroForm.

So a private extension is the right call, and `/DartPdfTextVerticalAlignment` as a `CosString` matches the existing convention (`DartPdfStampType`, `DartPdfStampTemplate` in `annotation.dart`). No objection to that half, and the README's interop caveat is accurate.

## Please reuse `PdfTextBoxVAlign` instead of `firstBaselineY`

`PdfTextBoxVAlign` (`text_box_appearance.dart`) already exists for precisely this and is threaded through the shared `writePdfTextBox`. The PR routes around it with a new placement path plus a `firstBaselineY` escape hatch that re-derives the geometry in `form_editor.dart`.

This isn't only conceptual duplication — two of the three new values are mathematically identical to what the enum already computes. Writing `L = lineHeight`, `s = fontSize`, `a = ascent`, `p = padding`, `H = box.height`, `n` lines:

- existing `centerBlock`: `bottom + (H + nL)/2 − (L−s)/2 − a` = `bottom + H/2 + (n−1)L/2 + s/2 − a`
- this PR's `center`: `top − p − (H − 2p − s − (n−1)L)/2 − a` = `bottom + H/2 + s/2 + (n−1)L/2 − a`

The padding cancels; they agree for every *n*. Likewise the new `top` reduces to `box.top − p − a`, which is `PdfTextBoxVAlign.top` verbatim. The only genuinely new placement is `bottom`, and the only real behavioural delta is the overflow clamp (`spare` floored at 0) — which `centerBlock` doesn't do and arguably should, since an overfull signature panel currently drifts its first line up under the clip.

Suggested shape:

- add `bottomBlock` to `PdfTextBoxVAlign`
- map `top → top`, `center → centerBlock`, `bottom → bottomBlock`
- keep the overflow clamp inside `writePdfTextBox` (behind a flag if you'd rather not change signature-panel output in this PR)
- drop `firstBaselineY` entirely

That preserves the contract the module's own header doc states — that the baseline ascent, alignment, wrap and clip logic "live, and are tested, in one place" — which the escape hatch quietly voids.

To be clear, keeping the public `PdfFormTextVerticalAlignment` enum is fine: form callers shouldn't have to reason about `centerLine` vs `centerBlock`. It's the placement math that shouldn't be duplicated, not the enum.

## Smaller notes

- `setTextFieldStyle(verticalAlignment: null)` can't clear, so callers need the separate `clearTextFieldVerticalAlignment`, which costs a second full appearance regeneration plus `_finishFieldEdit`. A `PdfFormTextVerticalAlignment.legacy` member (or a `clear: true` flag) would fold it into one pass.
- Auto-size uses `n * s * lineFactor` as its fit test while the new block height is `s + (n−1) * s * lineFactor`, so a `bottom`-aligned auto-sized block sits about `0.15s` above where the fit loop assumed. Not wrong, and there's a test pinning it, but worth a comment at the `blockHeight` line so nobody later "fixes" one to match the other.
- Unrelated formatting churn in the diff (the `PdfFieldType` switch reflow in `form.dart`, the `emit` line in `text_box_appearance.dart`) — presumably the formatter, harmless, just noise in review.

CI is green and the regressions cover the cases that matter (rotation, embedded fonts, multiple widgets, malformed metadata, flattening). Happy to merge once the placement math moves back into `text_box_appearance.dart`.

PR #906, opened 2026-09-10
Substitute metric-compatible faces for unembedded standard-14 text

ben-milanko:
## Summary

A capital `J` in base-14 Helvetica text drew with ~2.7pt of air after it — "Run JavaScript" reading as "Run J avaScript" on the example app's demo page (`/F1 12 Tf … (Run JavaScript) Tj`).

Nothing was wrong with the PDF or the placement machinery. A standard-14 font carries no font program, so the advances come from the AFM tables (`PdfFontInfo._fillStandardWidths`) while the glyphs came from whatever the host had — and `exactSubstitutedGlyphPlacement` (#649) pins each character to the PDF's own pen offset so selection, search and hit boxes land on the ink. DejaVu Sans, what the fallback chain reached on any platform without a real Helvetica (Linux, Windows, web, Android), draws `J` at 295/1000 em where Helvetica's table reserves 500, and `_buildPlacedLayout` spends that surplus as a cut — white space *inside* the word.

| glyph | Helvetica slot | DejaVu, as drawn | leftover |
| --- | --- | --- | --- |
| J | 500 | 275 | **+225** |
| S | 667 | 592 | +75 |
| R | 722 | 648 | +74 |
| P (next worst) | 667 | 603 | +64 |

This bundles URW's metric-compatible clones of the base 14 in the optional assets package and names them ahead of any host font: **TeX Gyre Heros** (Helvetica/Arial), **Termes** (Times), **Cursor** (Courier), joining the **Adventor** faces already shipped for Century Gothic. Regular, bold, italic and bold-italic each.

It is also less work per paint, not more — the same string through `pdfCanvas2dTextLayout`:

```
DejaVu metrics: [R@0.0, u@77.5, n@137.1, J@226.6, av@280.3, aS@393.6,
                 cr@524.8, i@614.2, p@638.0, t@697.7]   → 10 fillText calls
Heros metrics:  [Run@0.0, JavaScript@211.2]             →  2, kerning intact
```

API impact: `font_substitution.dart` (now `PdfBundledSubstitute` + `pdfBundledSubstituteFor`) is exported from `dart_pdf_editor`, because three renderers have to agree on the policy — the CanvasKit device, the canvas2d web worker (which now loads only the weights and slants a page actually shows, and none for an invisible OCR layer), and the flutter_gpu outliner (which reads the same bundled bytes and declines a run rather than outlining a face Canvas would not draw). `FlutterGpuSystemTextOutliner` gains a `ready` future for the one asynchronous moment it has. Without the assets package, unembedded standard-14 text now names the host's metric equivalents (Arial, Liberation Sans, Nimbus Sans and their serif/mono counterparts) before falling back further.

Times italics needed the widths as well: the base-14 fill mapped every `times*` name to `timesRomanWidths`, whose advances genuinely differ from the italic's (`A` is 611 against 722), so bundling Termes-Italic without that would have reproduced the same defect. The four tables already existed for `PdfStandardFont`; the fill now selects among them.

## Affected packages

- [ ] `pdf_cos`
- [ ] `pdf_document`
- [x] `pdf_graphics`
- [x] `dart_pdf_editor`
- [ ] `pdf_test_fixtures`
- [ ] Example app

Also `dart_pdf_editor_assets` (the fonts, their metric test, a new CI step) and `dart_pdf_editor_flutter_gpu`.

## Change type

- [x] Bug fix
- [ ] Feature
- [x] Rendering change
- [ ] Editing behavior change
- [x] Performance change
- [ ] Refactor / cleanup
- [ ] Tests / fixtures only
- [ ] Documentation only

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze` (packages + app, clean)
- [x] `cd packages/pdf_document && fvm dart test` (1039 pass)
- [x] `cd packages/pdf_graphics && fvm dart test` (1092 pass)
- [x] `cd packages/dart_pdf_editor && fvm flutter test` (2745 pass)
- [x] Ghent corpus test or baseline update — all 54 pass **with pixel comparison on**; no baseline change, since Ghent's patches embed their fonts
- [ ] Example app smoke test

Not run and why: `pdf_cos` is untouched. The local `corpus/` tree is git-ignored and not present in this environment. Also run and green: `dart_pdf_editor_flutter_gpu` (333, on Impeller + flutter_gpu), the new `dart_pdf_editor_assets` metrics suite (13), and the DartPDF app suite (574).

## User experience

- [x] Relevant keyboard, mouse, trackpad, touch, stylus, focus, and accessibility paths were checked — n/a, this changes only which face paints a run.
- [ ] Preview, commit, undo/redo, save, and reopen behavior agree where applicable.

On that last point, what the change costs is countable: **+1.46 MB** of declared font assets (**780 KB** gzipped over the wire, once, HTTP-cached) in apps that opt into the assets package, fetched and parsed by CanvasKit before the first frame, because the renderer names these families while painting. What it saves per paint is the piece count above.

`tool/perf.sh webdiff` cannot weigh those against each other — `app/tool/perf_harness` does not depend on `dart_pdf_editor_assets`, so its build carries neither the old nor the new fonts. The Patrol web E2E journey does drive the real app, but at three samples per scenario on a shared runner it cannot resolve a change of this size either: two runs of this branch whose only difference is a **test-only** commit reported `heavy-document` at −22.6% and then +17.2% against the same baseline, and `cad-image-deep-zoom` at −16.5% then +13.4%. Read the timings as noise in both directions. What is stable across both runs, and is the signal worth checking, is the structural side — identical scenarios, reconcile modes, fallback reasons, worker request kinds and transcript paths, and page-raster outcomes within one `miss`.

The macOS GPU corpus is the measurement that does hold still, and it reproduced exactly across two runs: acceptance unchanged (Ghent 49/8, PDF.js 178/1, no new coverage and no new Canvas fallbacks), with glyph draws 1104 → 1149 (+4.1%), stencil fan/cover −1.6% and geometry vertices −3.4% over the 227 common accelerated pages — the outliner resolving faces it used to decline.

The reproduction is the example app's own demo document — `_button(_jsButton, 'Run JavaScript')` in `demo_document.dart`, plain base-14 Helvetica in a form-button appearance stream. Rendered through the real pipeline at 4×, with the substitute face as the only variable, the page goes from `R un J avaScript` to `Run JavaScript`.

## Compatibility checklist

- [x] No `dart:io` imports in any `lib/` directory. (`system_text_outliner_native.dart` already had one; it is the native conditional-import half of an opt-in adapter, unchanged.)

## Notes for reviewers

- **The metric claim is checked, not assumed.** `packages/dart_pdf_editor_assets/test/substitute_metrics_test.dart` parses every shipped file with pdf_graphics' own OpenType parser and compares all 95 ASCII advances against `PdfStandardFont`'s AFM tables — Heros matches Helvetica and Helvetica-Bold exactly, Cursor is a flat 600, Termes matches all four Times tables. A new CI step runs it, so a font upgrade cannot quietly reintroduce the gaps.
- **The fonts ship unmodified** under the GUST Font License (`assets/fonts/TeXGyre-LICENSE.txt`, renamed from `Adventor-LICENSE.txt` since it now covers four families).
- **Cold start is a deliberate trade.** The five font-menu faces stay out of FontManifest because they preview rows most sessions never open; these are drawn on the first page that has text, so they are declared.
- **macOS rendering changes too**: Heros is now preferred over the system Helvetica. Same metrics, one consistent face across platforms.
- **The GPU corpus parity test needed the same faces.** `flutter_test` registers no pubspec font family and ignores `fontFamilyFallback`, so Canvas fell to the test box font while the outliner drew real outlines — mean pixel differences of 23 to 50 on the PDF.js pages with base-14 text. The test now registers the four bundled families from the same asset bytes the outliner reads, and settles that read before recording.
- **One wrinkle left alone**: code 39 is the single code where the checked-in AFM tables disagree with each other (StandardEncoding's `quoteright` vs WinAnsi's `quotesingle`; Times-Roman/Italic carry 333, Times-Bold/BoldItalic 278). The width fill is by code so it cannot serve both — the real fix is to fill by glyph name through the resolved encoding, which is a separate change. The metrics test accepts either glyph there and says why.
- Background: `doc/dev-log/2026-09-10-standard-14-metric-substitutes.md`.

https://claude.ai/code/session_01TNUc9iAc2jaM8drb5iXbRc

---
_Generated by [Claude Code](https://claude.ai/code/session_01TNUc9iAc2jaM8drb5iXbRc)_


ben-milanko:
`flutter-gpu-smoke` will finish red here, and the red is not this PR's — it fails the same way on `main`.

**What was mine, and is fixed.** On the first push the job died at step 7, *Validate native system-font GPU corpus parity*, with mean pixel differences of 23–50 on the PDF.js pages that carry base-14 text: `flutter_test` registers no pubspec font family and ignores `fontFamilyFallback`, so Canvas fell through to the test box font while the outliner drew real outlines. `gpu_corpus_test.dart` now registers the four bundled families from the same asset bytes the outliner reads, and settles that read before recording. On 453c401 that step passes, and steps 8–10 (the same-host `main` baseline and route comparison), which the failure had skipped, run clean.

**What remains.** Step 12, *Summarize Flutter GPU performance*, exits 65 with `Scenario … produced 0 complete run(s); required 6` for every scenario — the Patrol GPU journey emitted no `PdfPerfLog` events for it to summarize. The same step fails identically on `main`: run [34433883996](https://github.com/ben-milanko/dart-pdf/actions/runs/34433883996) (commit 8a48506) is red on `flutter-gpu-smoke` alone, with all eight other jobs green and its own perf comment reporting "No `PdfPerfLog` events were captured". The diff here touches font substitution and its tests; it does not touch the Patrol harness, the trace emitter, or the summarizer.

No fix for it exists to port, so I have not widened this PR with one, and I have no means to re-run the job from here. Everything else on the current head is green, `codecov/patch` included.

---

PR #909, opened 2026-09-10
Fix page-grid double-click landing on page 1

ben-milanko:
## Summary

Double-clicking a page in the full-area page grid (View options → page grid) closed the grid onto **page 1** instead of the page that was clicked.

The grid's own wiring was correct all along. `_PageTile._onTap` detects the second click, `_openPage` calls `viewerController.jumpToPage(index)` and fires `onOpenPage`, which the shell wires to `prefs.showThumbnailView = false`. Instrumenting `_scroll` showed the jump landing correctly while the grid was still up, and no later `jumpTo` at all — the offset was simply back at 0 on the first frame after the grid closed.

The cause was one level out, in `PdfShellPanelLayout.build`, which wrapped its result in `PdfPanelDragScope` / `PdfToolbarDragScope` only when the host had wired `onPanelDock` / `onToolbarDock`:

```dart
if (widget.onToolbarDock != null) {
  result = PdfToolbarDragScope(..., child: result);
}
```

Opening the grid sets `altView`, which hides the editing toolbar, which makes `PdfEditorView` pass `onToolbarDock: null`. So the wrapper dropped out of the tree on open and came back on close. A wrapper that comes and goes changes the depth of everything below it, so Flutter cannot match the old elements to the new ones: `_PdfViewerState` was discarded and rebuilt from scratch (a `print` in `initState` fired three times for one open/close round trip) and its `ScrollController` re-attached to a fresh `ScrollPosition` at `initialScrollOffset` — page 1. The viewer *was* still mounted throughout, as the existing test asserts; it just wasn't the same State.

Fix: both scopes now sit in the tree permanently and carry an `enabled` flag instead. `maybeOf` returns null when disabled, so `PdfToolbarMoveHandle` and a panel's move handle still hide exactly as before (that null check is the whole public contract of these scopes), and `updateShouldNotify` now fires on an `enabled` flip so a handle appears or disappears when docking is toggled. The tree shape no longer depends on whether the toolbar is showing.

The same wrapper churn would have rewound the viewer any other time the toolbar came and went; the grid is just where it was most visible, because that is the one place a click is *supposed* to move the viewer before the toolbar returns.

Background note: `doc/dev-log/2026-09-10-page-grid-double-click-page-1.md`.

## Affected packages

- [ ] `pdf_cos`
- [ ] `pdf_document`
- [ ] `pdf_graphics`
- [x] `dart_pdf_editor`
- [ ] `pdf_test_fixtures`
- [ ] Example app

## Change type

- [x] Bug fix
- [ ] Feature
- [ ] Rendering change
- [x] Editing behavior change
- [ ] Performance change
- [ ] Refactor / cleanup
- [ ] Tests / fixtures only
- [ ] Documentation only

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze` (no issues)
- [x] `cd packages/dart_pdf_editor && fvm flutter test` (2747 passing)
- [ ] Example app smoke test

If any relevant check was not run, explain why: the change is confined to Flutter widget-tree structure in `dart_pdf_editor`; it touches no parsing, rendering, or COS code, so the pure-Dart package suites and the render corpora have nothing to exercise here.

## User experience

- [x] Completion, cancellation, disabled, loading, and error states were considered. The `enabled: false` path is exactly the old "scope absent" behaviour: no move handle on the floating toolbar or on a docked panel.
- [x] Relevant keyboard, mouse, trackpad, touch, stylus, focus, and accessibility paths were checked. The grid's single-click/shift/⌘ selection behaviour is untouched; only the second click's destination changes.
- [x] Preview, commit, undo/redo, save, and reopen behavior agree where applicable — not applicable, no document state is involved.
- [x] Any journey-level latency, frame-time, or memory impact was measured or ruled out. Strictly less work than before: the viewer is no longer torn down and re-created (with a fresh page list, preview cache binding, and render worker sync) on every grid or toolbar toggle.

## Compatibility checklist

## Notes for reviewers

Regression test lives in `packages/dart_pdf_editor/test/pdf_shell_test.dart` — "a page grid double-click lands the viewer on that page": double-click Page 6 of an 8-page document in the grid, assert `viewerController.currentPage == 5` once it closes. It fails on `main` with `currentPage == 0`.

`PdfToolbarDragScope` and `PdfPanelDragScope` gained an optional `enabled` parameter that defaults to `true`, so any external construction of either is source-compatible.

https://claude.ai/code/session_01VWkdUBx5zFiPjtYy1ismH3

---
_Generated by [Claude Code](https://claude.ai/code/session_01VWkdUBx5zFiPjtYy1ismH3)_

### DIFF
--- a/doc/dev-log/2026-09-10-page-grid-double-click-page-1.md
+++ b/doc/dev-log/2026-09-10-page-grid-double-click-page-1.md
@@ -1,0 +1,48 @@
+# Double-clicking a page in the grid landed on page 1
+
+A double-click on a tile in the full-area page grid (`PdfThumbnailView`,
+View options → page grid) closed the grid but showed page 1 instead of the
+page that was clicked.
+
+The grid's own wiring was fine. `_PageTile._onTap` detects the second click,
+`_openPage` calls `viewerController.jumpToPage(index)` and then fires
+`onOpenPage`, which the shell wires to `prefs.showThumbnailView = false`. The
+jump really did happen - instrumenting `_scroll` showed the viewer sitting on
+the right page while the grid was still up, and no later `jumpTo` at all. The
+scroll offset was simply 0 again on the first frame after the grid closed.
+
+The cause was one frame further out, in `PdfShellPanelLayout.build`. It
+wrapped its result in `PdfPanelDragScope` / `PdfToolbarDragScope` only when
+the host had wired `onPanelDock` / `onToolbarDock`:
+
+```dart
+if (widget.onToolbarDock != null) {
+  result = PdfToolbarDragScope(..., child: result);
+}
+```
+
+Opening the grid sets `altView`, which hides the editing toolbar, which makes
+`PdfEditorView` pass `onToolbarDock: null`. So the wrapper vanished on open
+and came back on close. A wrapper that comes and goes changes the depth of
+everything below it, so Flutter cannot match the old elements to the new
+ones: `_PdfViewerState` was discarded and rebuilt from scratch each time (a
+`print` in `initState` fired three times for one open/close round trip), and
+its `ScrollController` re-attached to a fresh `ScrollPosition` at
+`initialScrollOffset` - page 1. The viewer *was* still mounted the whole
+time, as the existing test asserts, just not the same State.
+
+Fix: both scopes now always sit in the tree and carry an `enabled` flag
+instead. `maybeOf` returns null when disabled, so `PdfToolbarMoveHandle` and
+a panel's move handle still hide exactly as before (that null check is the
+whole public contract of these scopes), and `updateShouldNotify` now fires on
+an `enabled` flip so a handle appears or disappears when docking is turned on
+or off. The tree shape no longer depends on whether the toolbar is showing.
+
+The same wrapper churn would have rewound the viewer for any other reason the
+toolbar came and went, so this is not only a grid bug - it was just most
+visible there, because the grid is the one place where a click is *supposed*
+to move the viewer before the toolbar returns.
+
+Regression test: `pdf_shell_test.dart`, "a page grid double-click lands the
+viewer on that page" - double-click Page 6 of an 8-page document in the grid
+and assert `viewerController.currentPage == 5` after it closes.
--- a/packages/dart_pdf_editor/lib/src/editing/editing_panel.dart
+++ b/packages/dart_pdf_editor/lib/src/editing/editing_panel.dart
@@ -83,13 +83,15 @@ bool pdfPanelControlsRevealOnHover() => switch (defaultTargetPlatform) {
 
 /// Ambient wiring that lets a docked panel's move handle drive the shell's
 /// drag-to-redock affordance. [PdfShellPanelLayout] provides it around the
-/// panels; a panel's [PdfSidebarPanelGeometry.moveHandle] reads it. Absent
-/// when a shell wires no redocking - then panels render no move handle.
+/// panels; a panel's [PdfSidebarPanelGeometry.moveHandle] reads it. Reads as
+/// absent ([maybeOf] returns null) when a shell wires no redocking - then
+/// panels render no move handle.
 class PdfPanelDragScope extends InheritedWidget {
   const PdfPanelDragScope({
     super.key,
     required this.onDragStarted,
     required this.onDragEnded,
+    this.enabled = true,
     required super.child,
   });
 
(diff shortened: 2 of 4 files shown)


ben-milanko:
On the Patrol web performance comparison: I read those numbers as runner noise, not a regression from this diff. Flagging the reasoning so the red percentages aren't left unexplained for a reviewer.

The diff is four files: `shell_chrome.dart`, `editing_panel.dart`, a test, and a dev-log note. In runtime terms it adds exactly two `InheritedWidget` elements that are now permanently in the tree instead of conditionally, and changes `updateShouldNotify` from a constant `false` to a comparison of one bool that flips only when toolbar/panel docking is turned on or off. There is no path from that to decode, raster, or worker time.

What makes it look like machine noise rather than a real change:

- **Every scenario moved, including ones the diff cannot reach.** `jpeg-consumer-decode` +27.3% and `rounded-rung-pressure` +9.8% — the latter records no worker, no raster and no reconcile at all, yet its jank p95 still rose 23%.
- **The rise is uniform across unrelated signals** — jank +23%, reconcile +24%, raster +33%, worker total +26%, tile replay +18%. A code regression concentrates; a slower runner lifts everything together.
- **`jpeg-consumer-decode` contradicts itself**: elapsed p50 +27.3% while its own worker decode p95 went 2.0 ms → 1.0 ms. The decode got faster; the wall clock around it got slower.
- **Structural counts are stable**, which is what the harness says to actually watch: same scenarios and run counts, same reconcile modes (`fullReload` 1 / `incremental` 10 / `keyedStructure` 3), same fallback reason, same thumbnail paths, same prefetch batches. The only movements are `store` 19→20, `result-local` 14→15 and `result-worker` 97→99.
- **Three samples per cohort**, below the four the report itself needs before it will label an overlap as `within run spread`.

Directionally this change should if anything reduce work: hiding the toolbar previously tore down `_PdfViewerState` and rebuilt it — dropping its page list, preview bindings and render-worker wiring — and it no longer does.

To be straight about what this is: a mechanism argument plus the report's own structural signals, not a measurement. I have not re-run the Patrol journey. If you would rather see it confirmed, a re-run on the same commit should land back near the `main` baseline; if it comes back high again with structural counts still flat, that is worth a second look.

CI is otherwise green on `7c32f50` (`test`, `Android`, `iOS Simulator` and `flutter-gpu-smoke` still running at the time of writing), and the perf report is advisory — it does not gate.

---

ben-milanko:
Following up on my note above, since the Flutter GPU report answers the question it left open — and answers it better than the web one asked it.

That job is a **same-host A/B**: base `e06f211` and this PR alternated on one runner, 6 samples per cohort instead of 3. Under those conditions every one of the 44 scenario elapsed p50s comes back `within run spread`, and the GPU corpus structural comparison is identical across all 14 metrics — 7122 native draw calls, 14732 selected commands, 1798395 geometry vertices, +0.0% on each, over 227 common accelerated pages. No new GPU coverage, no new Canvas fallbacks. Jank total p95 is -3.0% and raster p95 +0.3%.

So when the baseline and the PR are measured on the same machine with enough samples to tell, this diff produces nothing measurable. That is the shape you would expect from two extra `InheritedWidget` elements, and it is the direct counterpart to the web report's 3-sample cross-run comparison against a separately-recorded `main` artifact.

I still have not re-run the Patrol *web* journey specifically, so that number stands unretested on its own terms — but I no longer think it is worth spending a run on.

---

ben-milanko:
Third perf report (native Android + iOS Simulator), and this one lets me replace my earlier inference with a checkable fact. Last word from me on perf.

**The Patrol journeys never trigger the code path this PR changes.** Every scenario setup pins both view toggles off:

```
perf_e2e_test.dart:47-48, 295-296, 570-571
  ..showThumbnailView = false
  ..showReflowView = false

demo_e2e_test.dart:20-21
  '…showReflowView': false,
  '…showThumbnailView': false,
```

`altView` is `reflowActive || gridActive`, so with both pinned false the editing toolbar is never suppressed, `onToolbarDock` never flips between null and non-null, and the wrapper that used to come and go never does. The diff contributes two extra `InheritedWidget` elements and nothing else — no rebuild avoided, none caused.

The native scenarios are further removed still: `native_perf_e2e_test.dart` pumps `MaterialApp(home: SizedBox())` and drives `PdfRetainedScene.record` and the tile raster backend directly. No `PdfEditorView`, no `PdfShellPanelLayout`, no `PdfViewer`. The changed classes are never constructed in `gpu-native-*` or `native-mobile-tiles`.

On the page-raster counts, which are the one thing here that looked structural rather than timing — Android `hit` 1→0 / `store` 3→1 / `miss` 103→115, iOS `hit` 1→0 / `store` 3→1: those are derived from what finished inside the measured window, and they moved the way a slower run moves them (fewer stores complete, so the single later hit becomes a miss). They are also the wrong direction for this diff, which keeps the viewer's caches alive across a toolbar toggle and so could only ever add hits.

Scoreboard across the three reports: the one methodologically sound job — Flutter GPU, same-host A/B, 6 samples — reports `within run spread` on all 44 scenarios and +0.0% on all 14 structural metrics. The three that show regressions (Patrol web, Android, iOS Simulator) all compare against a separately-recorded `main` artifact from a different runner.

I have not re-run any Patrol journey, and I don't intend to: the code under measurement is inert in all of them.

---

PR #911, opened 2026-09-11
Open exported pages after saving them

ben-milanko:
## Summary

Exporting pages out of a document - the thumbnail selection toolbar's export button, the page tile context menu's Export entry, and the page-range "Export pages…" dialog - wrote a standalone PDF and then said nothing at all. The user had just pulled pages out of a document and the only evidence was a file somewhere in the save dialog's folder. The app's `onExportPages` handler now opens what it wrote in a new tab.

All three entry points funnel through one host callback (`PdfEditorView.onExportPages`, fed by `exportSelectedPages()` / `exportPageRange()` / `exportPages(targets)`), so the change is entirely in `app/lib/editor_screen.dart`:

- `_exportPages` saves, then hands the same bytes to `_openBytes` - the path the OCR result and the initial document already take, so the new tab records a Recent and persists into the session like any other open document.
- The new tab adopts the save destination as its origin when the platform gives one (a desktop save dialog), with a fresh `securityBookmarkForPath`, so Save writes straight back to the file just created. A browser download or a mobile share sheet reports success with no path, so that tab opens over the exported bytes alone and its first Save asks for a location. Its title is the basename of where it landed, matching what Save As does to the tab it saved.
- The export path had no toast either, so a write that failed (full disk, folder gone read-only) passed in silence - `SaveResult.failed` carried a message nobody read. It now toasts `result.message` exactly like `_save`, which keeps a cancelled save quiet (null message).
- The source document is untouched: an export reads it, it never edits it.
- Cleanup: the `widget.saveDocumentAs ?? saveBytesAs(...)` lambda was copied out three times (`_save`, Reduce file size, and now the export). It is one `_saveAsDocument` helper, which is also what makes the export testable without the platform save dialog.

`PdfEditorView.onSplitPages` ("Split PDF…") still has no host handler in the app at all, so that action never appears - noted in the dev log as follow-up.

## Affected packages

- [ ] `pdf_cos`
- [ ] `pdf_document`
- [ ] `pdf_graphics`
- [ ] `dart_pdf_editor`
- [ ] `pdf_test_fixtures`
- [x] Example app
- [x] Documentation / CI / repository metadata only (dev-log note)

## Change type

- [ ] Bug fix
- [ ] Feature
- [ ] Rendering change
- [x] Editing behavior change
- [ ] Performance change
- [x] Refactor / cleanup
- [ ] Tests / fixtures only
- [ ] Documentation only

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze` (clean at the repo root)
- [ ] `cd packages/dart_pdf_editor && fvm flutter test`
- [ ] Example app smoke test

If any relevant check was not run, explain why: the change is confined to the app's host handler, so the app suite is the relevant one - `cd app && fvm flutter test` passes in full (590 tests), including the new `app/test/export_pages_test.dart`. Nothing under `packages/` or in the render path is touched, so the package suites and the corpus/Ghent sweeps have nothing to say about it.

## User experience

- [x] Completion, cancellation, disabled, loading, and error states were considered - the new test covers all four save outcomes: saved to a path (opens, toasts), cancelled (opens nothing, stays quiet), downloaded with no path (opens, toasts), failed (opens nothing, reports the failure).
- [ ] Relevant keyboard, mouse, trackpad, touch, stylus, focus, and accessibility paths were checked - no new surface; the existing export controls are unchanged.
- [x] Preview, commit, undo/redo, save, and reopen behavior agree where applicable - the exported tab opens clean (its bytes *are* its saved baseline), and with a path it saves in place rather than re-prompting.
- [x] Any journey-level latency, frame-time, or memory impact was measured or ruled out - an export already held the extracted bytes; opening them is the same work any `_openBytes` does, and nothing in the render or interpreter path changed.

Programmatic only: the test builds a 4-page fixture (`buildMultiPagePdf(4)`) and asserts the extracted pages with `PdfTextExtractor` over the new tab's document, rather than just counting tabs.

## Compatibility checklist

## Notes for reviewers

One judgement call worth confirming: "open" here means open in a new DartPDF tab, not hand the file to the OS viewer. That is the reading that matches the app (it is a PDF editor, and `_openBytes` already exists for the OCR result), and it works identically on desktop, web and mobile where a reveal-in-file-manager would only work on desktop. Say the word if you wanted the system handler instead.

No new localized strings: the toast reuses the existing `SaveResult` messages.

Follow-up noted in `doc/dev-log/2026-09-11-open-exported-pages.md`: Split PDF… is unreachable in the app because `onSplitPages` is unwired; whatever opens its outputs will want this same treatment.

https://claude.ai/code/session_01HfAaQn8chm858rb6ohFkVk

---
_Generated by [Claude Code](https://claude.ai/code/session_01HfAaQn8chm858rb6ohFkVk)_

### DIFF
--- a/app/lib/editor_screen.dart
+++ b/app/lib/editor_screen.dart
@@ -2510,9 +2510,6 @@ class _EditorScreenState extends State<EditorScreen>
   Future<void> _save(DocumentTab tab, {bool saveAs = false}) async {
     final bytes = tab.session?.bytes;
     if (bytes == null) return;
-    final saveAsDocument = widget.saveDocumentAs ??
-        (ctx, bytes, name) =>
-            saveBytesAs(ctx, bytes, name, pdfLabel: appL10n(ctx).fileTypePdf);
     final saveToPath = widget.saveDocumentToPath ?? saveBytesToPath;
     final inPlace = !saveAs && tab.originPath != null && supportsInPlaceSave;
     var result = inPlace
@@ -2521,11 +2518,11 @@ class _EditorScreenState extends State<EditorScreen>
             tab.originPath!,
             bookmark: tab.originBookmark,
           )
-        : await saveAsDocument(context, bytes, tab.title);
+        : await _saveAsDocument(context, bytes, tab.title);
     if (!mounted) return;
     if (inPlace && !result.succeeded) {
       // The origin couldn't be written (moved, read-only) - offer save-as.
-      result = await saveAsDocument(context, bytes, tab.title);
+      result = await _saveAsDocument(context, bytes, tab.title);
       if (!mounted) return;
     }
     if (result.succeeded) {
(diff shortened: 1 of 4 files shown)


ben-milanko:
The `test` check failed on 663f2be, and it isn't this PR's:

```
pdf_shell_test.dart: PdfEditorView a page grid double-click lands the viewer on that page
The finder "Found 0 widgets with key [<'pdf-shell-page-grid'>]" (used in a call to "tap()") could not find any matching widgets.
2752 tests passed, 1 failed, 33 skipped.
```

The same single failure is red on the base commit 12f5b49 ([job](https://github.com/ben-milanko/dart-pdf/actions/runs/34572260127/job/103176721191)), so main is red for every PR off it. This diff is confined to `app/`, which a `packages/dart_pdf_editor` widget test can't see.

Cause: #908 replaced the loose view-mode checkmarks with one `SegmentedButton` keyed `pdf-shell-view-mode`, and added the file's own `tapViewMode(tester, label)` helper because a segment is addressed by its label rather than a per-item key. Every other test in the file was converted; this one still tapped `ValueKey('pdf-shell-page-grid')`, which now belongs to the *grid widget itself* (`pdf_editor_view.dart`) and isn't in the tree until the mode is already on — hence zero matches. (`pdf-shell-page-grid-toggle` isn't it either: that key is on the compact Controls sheet's tile, which the desktop popup doesn't build.)

Fixed in cbcdb43 — one line, `await tapViewMode(tester, 'Page grid');` — ported into this branch so it can reach green rather than waiting on a separate PR; it no-ops once main carries the same change. Validated locally: `cd packages/dart_pdf_editor && fvm flutter test` → 2753 passed, 33 skipped (CI's own count with that one failure flipped), plus `dart analyze` clean and the full `app` suite still green.

---

PR #916, opened 2026-09-16
Render the Linux app with Skia so PDF text is antialiased (#912)

### LINKED ISSUE #912: [Bug]: test stroke issue
### Affected package

example app

### What happened?

the same pdf doesnot look clear in dart pdf but clear and vibrant in okular

Image

Image

### Minimal reproduction

open same pdf in okular linux and dart pdf linux

### Error output or stack trace

```text

```

### Version or commit

lastest

### Platform

Flutter Linux

### Environment details

```text

```

### PDF details

_No response_

### Checks already run

```shell

```

### Before submitting

- [x] I searched existing issues and did not find a duplicate.
- [x] I can provide a non-confidential reproducer or enough detail to create one.

ben-milanko:
Fixes #912

## Root cause (verified)

The reporter's file is the widely shared 675-page Acrobat merge of *Operating Systems: Three Easy Pieces*. Its page 1 content stream is empty. The title is two `/FreeText` typewriter annotations drawn with an embedded TrueType BaskervilleOldFace, filled with plain `Tj` and no stroke or text render mode tricks.

- **Our renderer is correct.** `render_smoke_test.dart` on that page (direct, render worker, and `--enable-impeller` on Metal) matches `pdftoppm` at the same scale.
- **The report shows zero antialiasing.** The title in the Linux screenshot has exactly **two** grey levels, 0 and 255. The app's own UI text in the same screenshot has more than 80, and Metal Impeller renders the title with five (4x MSAA). So page vector content went through a single-sample rasterizer.
- **The engine explains why (Flutter 3.47 source):**
  - `fl_dart_project.cc` sets `enable_impeller = TRUE` by default, and the Linux embedder always uses OpenGL.
  - `impeller/renderer/backend/gles/capabilities_gles.cc` reports offscreen MSAA only for `GL_EXT_multisampled_render_to_texture2` or an ES 3+ context. Without either, `Picture.toImage` snapshots are single-sample.
  - Impeller has no analytic AA for arbitrary path fills; its SDF paths cover rects, rrects, circles and lines. PDF glyphs are outline paths, so they come out as 1-bit masks and thin serifs and hairlines drop out (the T turning into `I'`). The glyph atlas behind Flutter's own text is unaffected.
  - The flutter_gpu tile backend isn't involved: it already rejects GLES contexts.

## Fix

- `app/linux/runner/my_application.cc` calls `fl_dart_project_set_enable_impeller(project, FALSE)` before the headless windowing engine or view is created. Skia antialiases paths without MSAA.
  - `DARTPDF_IMPELLER=1` turns Impeller back on for A/B testing. `FLUTTER_ENGINE_SWITCHES` is ignored in release builds, so the app needs its own switch.
- The example runner gets the same call.
- `app/test/linux_runner_renderer_test.dart` pins the call in both runners, so a template regeneration can't silently drop it.
- A CHANGELOG entry, plus a dev-log note: `doc/dev-log/2026-09-16-linux-impeller-aliased-paths.md`.

No Dart rendering code changed, so Ghent baselines are unaffected.

## Perf / trade-offs

Linux only. The Impeller-only paths (strip shaders gated on `ui.ImageFilter.isShaderFilterSupported`, flutter_gpu tiles) turn off on Linux. It goes back to the Canvas + flat-replay path it used before Impeller became the Linux default. flutter_gpu was already rejected on GL. Not measured on a Linux desktop.

## Not verified here

This was developed on macOS. I did not run the Linux build, so it needs a check on a Linux desktop (for example `DARTPDF_IMPELLER=1` vs the default) before release. Other Linux hosts that embed `dart_pdf_editor` need the same runner call.


sudo-muneeb:
> This was developed on macOS. I did not run the Linux build, so it needs a check on a Linux desktop (for example DARTPDF_IMPELLER=1 vs the default) before release. Other Linux hosts that embed dart_pdf_editor need the same runner call




the problem is fixed, FYI

PR #918, opened 2026-09-17
Give each window its own view mode

ben-milanko:
## Summary

Switching to the page grid in one window switched every other window with it.

The app owns exactly one `PdfEditingPreferences` and hands it to every window - deliberately, so tool styles, panel layout, theme and viewport memory are the same wherever you look (`DartPdfEditorApp._prefs`). The view mode lived there too, and every shell read it live (`prefs.showThumbnailView`), so the page grid was not a window's view of its document but a process-wide setting with a UI that looked local: pick it in one window and every other window dropped its page out from under the reader.

That the mode persists is right; that it is shared while you are looking at it is not. Two windows are two reading surfaces, and the surface owns its mode.

**API impact** (`dart_pdf_editor`):

- `PdfViewModeHolder` - the two-member surface the chrome binds to (`viewMode` + `Listenable`). `PdfEditingPreferences` already had both and implements it as-is.
- `PdfViewModeController` - one window's live mode. Each change writes through to the preferences, so the value still persists, now as the mode the NEXT window starts in rather than an order to the windows already open.
- `PdfEditorView(viewMode:)` / `PdfReader(viewMode:)` take a holder. A shell given none owns a controller seeded from - and written back to - its preferences, which is exactly the behaviour a single-window host had, so embedders are unchanged.
- `pdfShellViewModeControls` takes `viewMode:` in place of `preferences:`; `PdfShellViewOptionsButton` gains an optional `viewMode:` that falls back to its preferences. Neither is exported.

App: `EditorScreen` (one per window) owns the controller and passes it to every shell that window builds and to its own palette commands, so all of a window's tabs share one mode and no other window follows.

Preferences load asynchronously and the first window is built before the store answers, so a controller adopts the stored mode when `ready` completes - unless the user has already picked one in that window, which wins for the same reason `PdfEditingPreferences._modified` lets a value set during the read beat the stored one.

Background: [doc/dev-log/2026-09-17-per-window-view-mode.md](doc/dev-log/2026-09-17-per-window-view-mode.md).

## Affected packages

- [ ] `pdf_cos`
- [ ] `pdf_document`
- [ ] `pdf_graphics`
- [x] `dart_pdf_editor`
- [ ] `pdf_test_fixtures`
- [x] Example app
- [x] Documentation / CI / repository metadata only (dev-log + changelogs)

## Change type

- [x] Bug fix
- [ ] Feature
- [ ] Rendering change
- [x] Editing behavior change
- [ ] Performance change
- [ ] Refactor / cleanup
- [ ] Tests / fixtures only
- [ ] Documentation only

## Validation

- [x] `fvm flutter pub get`
- [x] `fvm dart analyze` (no issues)
- [x] `cd packages/dart_pdf_editor && fvm flutter test` (2797 passed)
- [ ] Example app smoke test

Also ran the app suite: `cd app && fvm flutter test` - 594 passed.

The unchecked boxes are the pure-Dart rendering/parsing suites and the corpus sweeps: nothing below `dart_pdf_editor` is touched (no parser, renderer or COS code in the diff), so there is no path for this to change any rendered or parsed output.

## User experience

- [x] Completion, cancellation, disabled, loading, and error states were considered.

Every surface that picks a mode goes through the same holder: the desktop view-options `SegmentedButton`, the compact Controls sheet tiles, the app's command palette, and the page grid's own "open this page" (which now asks for `PdfViewMode.pages` rather than clearing one flag). Reopening is unchanged - the mode is still persisted, and a new window or a fresh launch starts in the last one picked. No render, cache or layout path is touched, so there is no latency or memory surface: the change is which object the chrome reads a mode from.

## Compatibility checklist

## Notes for reviewers

Tests cover both halves of "per window":

- `editing_preferences_test.dart` - one window's choice leaves another's alone while still persisting for the next window; the late-load adoption; a mode picked during the read beating the stored one.
- `pdf_shell_test.dart` - two `PdfEditorView`s over one preferences object, each with its own controller; the grid appears in the first window's subtree only.
- `multi_window_test.dart` - the same at app level with two `EditorScreen`s, plus the complement: after another window writes a different mode to the shared preferences, this window's next tab still opens into the mode this window is in. That second test is the one that fails if the app wiring is dropped and each shell seeds itself from the preferences - per-tab looks right until another window moves.

The reflow view rode on the same shared flag and moves with the page grid; keeping one per-window and the other process-wide would have split an invariant that exists precisely to keep the two coherent.

https://claude.ai/code/session_013sjWdQ7WMwN85va3tRTVhi

---
_Generated by [Claude Code](https://claude.ai/code/session_013sjWdQ7WMwN85va3tRTVhi)_


ben-milanko:
On the Patrol web performance headline (+10% to +32% across the scenarios): I don't think any of it is attributable to this PR, and I'm not chasing it here.

The diff touches five files — `editing_preferences.dart` (adds `PdfViewModeHolder`/`PdfViewModeController`), `pdf_editor_view.dart`, `pdf_reader.dart`, `shell_chrome.dart`, `editor_screen.dart` — and nothing in the render, decode, tile, worker or cache paths. The only per-frame difference is one extra `Listenable` in the shells' `ListenableBuilder.merge`; the `gridActive` / `reflowActive` computation is the same expression read off a different object.

What the scenarios measure is exactly the code that isn't touched, and the numbers move as a block rather than in one place:

- `jpeg-consumer-decode` +32.2% is image decode with no view-mode surface at all, and `cad-image-deep-zoom` decode p95 +40.7% is the same story.
- `rounded-rung-pressure` reports elapsed +9.2% while its jank p95 *improves* 82% — opposite directions in one scenario.
- The structural counts a real change would move are flat: identical scenarios, reconcile modes, fallback reasons and thumbnail paths; page-raster outcomes differ by 1-2. Tile replay requests (375 → 300) and prefetch batches (36 → 20) fall out of a slower run doing less idle prefetch, not out of a different tiling decision.

Each cohort is 3 samples, below the 4 the harness needs before it will label an overlap `within run spread`, so the percentages are raw p50 deltas on shared runners. A uniform slowdown across unrelated workloads is what runner variance looks like.

The `Patrol performance report` check itself is green and the comparison is advisory. Happy to re-run the Patrol job for a second sample if you'd rather see it confirmed than argued.

---

ben-milanko:
Following up on my note about the web perf headline: the Flutter GPU report is the confirming sample, and it lands the other way.

It runs a same-host A/B with 6 samples per cohort, enough for the harness to apply its interquartile test — and every scenario but one comes back `within run spread`, with jank p95 -23.1% and raster p95 -0.8%. The corpus and structural counts are exactly equal to `main`: Ghent 49/8, PDF.js 178/1, no new GPU coverage, no new Canvas fallbacks, and every draw metric at +0.0% down to the geometry vertex count (1798395 on both).

Same diff, same runners, 6 samples instead of 3 — and the differences disappear. No re-run needed unless you'd still like one.

---

ben-milanko:
One correction to read the native report by: the iOS half of it came from a run that was cancelled mid-flight, so those percentages aren't a comparison of anything.

Pushing the follow-up commit at 12:14 cancelled the in-flight run for `1fa5d7e` ([iOS Simulator job](https://github.com/ben-milanko/dart-pdf/actions/runs/35217912939/job/105190694377): `Run Patrol on iOS` cancelled, `Collect Patrol iOS performance trace` failed). That is exactly what the table shows — 4 samples instead of 6, 13 journey segments instead of 17, 88 prefetch batches instead of 132 — and a truncated journey is what the +38.9% / +61.3% entries are measuring. The Android half of the same report ran to completion with 6/6 samples and is `within run spread` on every scenario, with jank p95 -4.8% and raster p95 -1.0%.

iOS re-runs from scratch on the new head, so there will be a complete cohort to look at shortly. Nothing to do about the cancelled job itself — it was superseded, not broken.

---

ben-milanko:
The web report re-ran on `9a4a6d4` with much the same percentages, so two runs agreeing is worth explaining rather than repeating myself. It is not two confirmations: both compare against the *same stored* `main` artifact, so a baseline that happens to be fast shows up as a regression every time.

That baseline moves a lot. #917 — a vertex-reshape change that touches the decode, worker and tile paths exactly as little as this one does — got the mirror image of these numbers against its own `main` artifact:

| Scenario | #917 (main → PR) | #918 (main → PR) |
| --- | --- | --- |
| `cad-image-deep-zoom` | 13962 → 10346 ms (**-25.9%**) | 12902 → 14648 ms (**+13.5%**) |
| `heavy-document` | 9963 → 6338 ms (**-36.4%**) | 8180 → 9895 ms (**+21.0%**) |
| `jpeg-consumer-decode` | 1591 → 1026 ms (**-35.5%**) | 1180 → 1428 ms (**+21.0%**) |
| `rounded-rung-pressure` | 746 → 675 ms (**-9.5%**) | 682 → 738 ms (**+8.2%**) |
| `web-worker` | 2329 → 1640 ms (**-29.6%**) | 1853 → 2306 ms (**+24.4%**) |

Look at the `main` columns: 9963 ms vs 8180 ms for the same scenario on the same branch, 1591 vs 1180 on another. The baseline alone varies by 20-35% between artifacts, and the whole five-scenario block moves together in whichever direction the stored artifact happens to sit. #917 did not make the web 30% faster and this PR has not made it 20% slower.

The methodological difference is visible in the reports themselves: the Flutter GPU harness alternates base and PR **on one runner** and found nothing (every scenario `within run spread`, all draw counts identical). The web harness compares a fresh run against a stored artifact from a different runner, with 3 samples and no interquartile test. That is the gap these percentages live in.

So I'd read the web headline as uninformative for this PR rather than as a signal to chase — but it is the harness's own comparison, not mine to change here.

---

